PR #166, opened 2020-04-16
Mccalluc/pretty enum

### LINKED ISSUE #162: enums in docs don't make sense to people
Probably something like
> `A1` or `B2` or `C3`

would be more clear.

mccalluc:
Tiffany: If you can do a code review, and look at the [new docs](https://github.com/hubmapconsortium/ingest-validation-tools/blob/3175beacb3ecb2eb14cb2f26f1644a5247442e1c/docs/atacseq/README.md#assay_category): Do you think these are more clear? Or do you think another way of presenting the information would be better? Fix #162.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix `--type_metadata` so it still works without a submission directory.
 ### Changed
 - Make the ATACseq validation more flexible.
+- Less confusing representation of enums in docs.
 
 ## [v0.0.1](https://github.com/hubmapconsortium/ingest-validation-tools/tree/v0.0.1) - 2020-04-13
 ### Added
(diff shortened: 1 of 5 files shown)


tsliaw:
I think the repeating "or" is a little distracting for the docs and I don't know if it is any more clearer than what the original docs had. Do you think just having one "or" before the last item would suffice and having commas separate the other items? For example: 
"whole cell, nucleus, cell-cell multimer or spatially encoded cell barcoding"

Because having a lot of "or" when there is a lot of items makes it harder to read each individual item, but when there was only two items, for example "none or FACS", I personally thought that was clearer to read.

mccalluc:
Good feedback: What about...?
> `red`, `green`, `blue`

The commas are less distracting, and the `<code>` formatting of each item also helps to mark it off.

mccalluc:
actually, let me do the final "or" too.

tsliaw:
I think your new changes are a lot more clearer.

PR #174, opened 2020-04-22
Mccalluc/doi required

### LINKED ISSUE #90: make protocols DOIs required again.

mccalluc:
Fix #90.

@cebriggs7135 : Can you confirm that now that you have a command-line option, all of the DOIs should be considered as required?

@john-conroy: There's not much code to review, but sometimes seeing things at a small scale is easier than looking at it from the top down.


john-conroy:
The constraints listed in the tables-schema yaml files were removed outright. Should they not instead be changed to true?

cebriggs7135:
All protocols fields should be populated.

mccalluc:
@john-conroy :
> The constraints listed in the table-schemas yaml files were removed outright. Should they not instead be changed to true?

The default semantics for Table Schema are for fields to be optional... but since nearly all of our fields are required, [table_schema_loader](https://github.com/hubmapconsortium/ingest-validation-tools/blob/master/src/ingest_validation_tools/table_schema_loader.py) reads in the yaml, combines separate files together, and then adds additional constraints before using it for validation. This makes the yaml as easy to author as possible... at the cost of departing from the usual table-schema semantics.

PR #175, opened 2020-04-22
Mccalluc/ignore dot files

### LINKED ISSUE #173: Ignore invisible files
Anything that comes from a Mac is likely to have `.DS_Store` files. I propose ignoring these files by default, but leaving strict validation as an option. Ok?

mccalluc:
Fix #173.

@cebriggs7135 : This will prevent stray invisible files whose names begin with `.` from causing an invalid submission. Mostly commonly, this is `.DS_Store` on Macs. Perhaps is should be more strict, and only ignore that particular name? For now, when we're trying to get _anything_ to validate, I think this is ok. Does it have your approval?

@john-conroy : Code review. Thanks!


cebriggs7135:
Ignore them.

PR #176, opened 2020-04-22
Mccalluc/ignore files cli option

### LINKED ISSUE #169: Option to ignore metadata files that have been dropped into directory.
This would allow Joel to validate the directories they've been populating.

mccalluc:
Fix #169

This is a workaround for what is really a temporary situation, so if there were a lighter-weight solution, I'd go for it. In the long run the metadata.tsv should be _outside_ the submission directory... but for now it's on the _inside_, because there's not a good place to put it at a higher level.


mccalluc:
@john-conroy : When you have chance to look at this. Thanks!

Issue #163, opened 2020-04-16
Validate that types outside match types inside.


mccalluc:

mccalluc:
Not automatic, but type schemas can now override level 1. closing.

PR #177, opened 2020-04-23
Mccalluc/inside outside types

mccalluc:
@cebriggs7135 : The idea is that type-specific schemas can override more generic definitions from Level 1. See the [change in the docs below](https://github.com/hubmapconsortium/ingest-validation-tools/compare/mccalluc/inside-outside-types?expand=1#diff-04688e4d986f44eb995aad5f191b874c)... Does this seem like the right course, and are the particular changes here accurate?

@john-conroy : Code review. Thanks!

Towards #163... but will need to be applied to the other types as well.


mccalluc:
@cebriggs7135 : I think we talked this one through, but if it's good, can you review and approve this?

mccalluc:
and @john-conroy : Can you code-review? Thanks!

PR #229, opened 2020-05-07
modify atacseq constraints for cell_barcode_read/offset/size

cebriggs7135:
These should be required: True but with no required pattern, just like scrnaseq.

### DIFF
--- a/src/ingest_validation_tools/table-schemas/level-2/atacseq.yaml
+++ b/src/ingest_validation_tools/table-schemas/level-2/atacseq.yaml
@@ -109,17 +109,17 @@ fields:
     name: cell_barcode_read
     description: Which read file contains the cell barcode.
     constraints:
-      pattern: '[A-Z]\d+(,[A-Z]\d+)*'
+      required: True
   -
     name: cell_barcode_offset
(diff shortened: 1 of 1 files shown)


mccalluc:
@cebriggs7135 :
- `required: True` is the default and does not need to be specified explicitly: See [generated docs](https://github.com/hubmapconsortium/ingest-validation-tools/tree/master/docs/atacseq#cell_barcode_offset).
- Can you clarify why `offset` and `size` need to unconstrained? This worries me.

Talking might be easier than typing.

cebriggs7135:
Because Stanford represents barcode_read in their recent submission as:
R2 (i5)
meaning the barcode is in Read#2 and includes an index.

cebriggs7135:
Your link to generated docs brings me to a page with nothing much: 
Screen Shot 2020-05-10 at 4 17 05 PM

mccalluc:
URL changed when "sc" was prepended: https://github.com/hubmapconsortium/ingest-validation-tools/tree/master/docs/scatacseq#cell_barcode_offset

I can address `cell_barcode_read`: For the others, the values we collect need to be meaningful to anyone in the field, not just a reflection of one lab's practices.

mccalluc:
@cebriggs7135 : I've addressed part of this, and I'm uncomfortable with the part that remains.

>  No one will ever search for lab-specific abbreviations for reads so it's not important.

Is your belief that the information will be useful in context, even if it's not uniform between labs?

I'd suggest closing this PR and opening new small PRs for the issues which remain.

PR #255, opened 2020-05-13
Make description require the RIN value for RNA quality

cebriggs7135:
Edit field in bulkrnaseq:
bulk_rna_isolation_quality_metric_value
to make RIN value required.


mccalluc:
@cebriggs7135 : So this will be a number now instead of a descriptive string?

cebriggs7135:
Yes. RIN# will be required It ranges from 1-10.

PR #257, opened 2020-05-15
Add a column to LC-MS to describe combined tissue_id slices

cebriggs7135:
These. are currently denoted as  "#_#" appended to the tissue_id, indicating a range of tissue slices.

This field may be added to other metadata.tsv schema.

### DIFF
--- a/src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
+++ b/src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
@@ -10,6 +10,11 @@ fields:
 
   # Level 2 fields:
   -
+    name: combined_tissue_sections
+    description: If a number of tissue sections were combined for this assay, provide a comma-delimited list of the tissue
+    slice ids or #:# for contiguous slices. These sections may be denoted as "#_#" appended to the tissue_id, indicating a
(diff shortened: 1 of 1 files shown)


mccalluc:
@cebriggs7135 : What are the IDs that would be used here? I think we want to avoid making on-going references to the sequential integer IDs, but I think that's what would happen here. Also not sure if this belongs down at level 2 if it's really about provenance?

Should probably ping @ngehlenborg .

cebriggs7135:
Ajay requests a column to tell the viewer that the assay was done on combined tissue slices. The slices are a known distance apart so this helps the user interpret how thick the tissue sample was. I put it in level 2 because I don't know if it is relevant to more than a few assays.

mccalluc:
Ok, that's a good usecase, but I'm pretty sure it can be derived from what we know about provenance, at least as I understand it... I don't think it's something we need to ask the data submitters to provide. If you still believe this should be merged, let's loop in Nils.

cebriggs7135:
We'll discuss at our curation mtg on Friday...

cebriggs7135:
This issue is resolved with an email to Ajay explaining why tissue_ids are never seen by the user.

PR #270, opened 2020-05-18
WGS: dna_assay_input to dna_assay_input_value

cebriggs7135:
Also changed dna_assay_input_units enum to ug.


src/ingest_validation_tools/table-schemas/level-2/wgs.yaml
```diff
@@ -12,14 +12,13 @@ fields:
...
-- name: dna_assay_input
+- name: dna_assay_input_value
   description: Amount of DNA input into library preparation
   type: number
 - name: dna_assay_input_unit
   description: Units of DNA input into library preparation
   constraints:
-    enum:
-      - nuclei
+  enum: ug
```
mccalluc, line 21: @cebriggs7135 : In this file, indenting matters -- It doesn't know that `enum` belongs to `constraints` unless it's indented. I'll fix -- no worries.

mccalluc: ... and since we want a list of options, even if there's only one, need to keep the dash.

PR #272, opened 2020-05-18
LCMS: 3 changes--um to micron; A to a ; mass range unit m/z

cebriggs7135:
Added mass_range_unit to m/z.
Changed lc_id_unit enum: from um to micron
Changed the following fieldnames:
    lc_mobile_phase_A to lc_mobile_phase_a
    lc_mobile_phase_B to lc_mobile_phase_b

### DIFF
--- a/src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
+++ b/src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
@@ -28,6 +28,11 @@ fields:
     name: mass_range_high_value
     description: The high value of the scanned mass range for MS1. (unitless)
   -
+    name: mass_range_unit
+    description: Units of the mass range values
+      enun
+       - m/z
+  -
     name: data_collection_mode
     description: Mode of data collection in tandem MS assays. Either DDA (Data-dependent acquisition) or DIA (Data-indemendent acquisition.
   -
@@ -80,7 +85,7 @@ fields:
     description: units of LC column inner diameter (typically microns)
     constraints:
       enum:
-        - um
+        - micron
   -
     name: lc_flow_rate_value
     description: Value of flow rate.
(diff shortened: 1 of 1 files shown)


src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
```diff
@@ -27,6 +27,11 @@ fields:
   -
     name: mass_range_high_value
     description: The high value of the scanned mass range for MS1. (unitless)
+  -
+    name: mass_range_unit
+    description: Units of the mass range values
+      enun
+       - m/z
```
mccalluc, line 34: - mass range is literally unitless.

src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
```diff
@@ -80,7 +85,7 @@ fields:
     description: units of LC column inner diameter (typically microns)
     constraints:
       enum:
-        - um
+        - micron
```
mccalluc, line 88: Unless one community strongly prefers micron to um, I think we should be consistent?

src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
```diff
@@ -97,10 +102,10 @@ fields:
     name: lc_gradient
     description: LC gradient
   -
-    name: lc_mobile_phase_A
+    name: lc_mobile_phase_a
     description: Composition of mobile phase A
   -
-    name: lc_mobile_phase_B
+    name: lc_mobile_phase_b
```
mccalluc, line 108: I'll make a separate PR with these.

src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
cebriggs7135: Yes. It is a ratio of the mass value divided by the charge value. m/z describes how the mass is calculated. mass spectrometry output is called "m/z", not just "mass". z is usually 1 but not always.

src/ingest_validation_tools/table-schemas/level-2/lcms.yaml
cebriggs7135: "micron" is the standard, I am discovering. If you keep, then I will have to change & resubmit a number of metadata.tsvs.

PR #278, opened 2020-05-22
Mccalluc/link to datatype docs

mccalluc:
@cebriggs7135 : Spot check these and approve if these links are going to the right documents.


cebriggs7135:
Only 3 or 4 have active links that take me to a document. The rest generate an error:
Screenshot_20200523-195250

mccalluc:
@cebriggs7135 : 
- When reporting something that doesn't work, a url is useful: In your screenshot there's not enough visible to tell where you are.
- When you include a URL, please include it as text.
- The 404 you are getting is from github pages, while the new urls are all to google docs, so I'll really not sure what you're looking at.
- If you want to look at the rendered documents, [this link](https://github.com/hubmapconsortium/ingest-validation-tools/tree/dda166c5dd64543f76559127a8b4375b1b2c73c2/docs) should work.

cebriggs7135:
Ok. I'll get back to this tomorrow & do what you suggest above.

mccalluc:
@cebriggs7135 : 
- Please don't re-write comments, beyond just correcting typos. It makes it harder for someone in the future to understand the thread of comments.
- Right, the links on master are bad right now (they are just "TODO"), and this PR fixes them.
- If you approve this change, please click the button to `Review`, and then select `Approve`.

Thanks!

PR #281, opened 2020-05-24
Mccalluc/easier directory validation

mccalluc:
Stepping back from it, I realized that using json schema to validate directory structure was not really sustainable. Though it gives you a lot of flexibility, the schemas were verbose and hard to write, and the errors are verbose and hard to read.

This replaces it with a system based on describing the directory structure in nested dicts, with globs as the keys. For autoflourescense, Vanderbilt describes the structure like this:

<details>

```
DR_reformat….
    /LC Data
    /metadata
        /assay
            AF.csv
/templates
/VAN000*-*K-**-**_**
    /CCF Metadata
        VAN000*-*K-**-**_ccf_metadata
    / VAN000*-*K-**-**_**-AF
        /processedMicroscopy
            /VAN000*-*K-**-**-AF_preIMS_images
                /VAN000*-*K-**-**-AF_preIMS_registered.ome.tiff
            /VAN000*-*K-**-**-AF_preIMS_transformations
                /transform_00_VAN000*-*K-**-**-AF_preIMS_rigid.txt
                /transform_01_VAN000*-*K-**-**-AF_preIMS_affine.txt
/rawMicroscopy
    /VAN000*-*K-**-**-AF_preIMS_meta_unregistered.xml
    /VAN000*-*K-**-**-AF_preIMS_unregistered.czi
``` 

</details>

Going forward, I'll be able to take something like this, add some colons, and that will just about do it.

One question: I'm using globs rather than regex... I could switch up, but my sense is that catching the big structural problems is enough, and that the primary audience for this doesn't really understand regex, in any case.


src/ingest_validation_tools/directory_validator.py
```diff
@@ -0,0 +1,96 @@
...
+    '''
+    required_globs, allowed_globs = _get_required_allowed(paths_dict)
+    required_missing_errors, not_allowed_errors = ([], [])
+    if not path.exists():
+        raise DirectoryValidationErrors({
+            'No such file or directory': str(path)
+        })
+    actual_paths = []
+    for triple in walk(path):
+        (dir_path, dir_names, file_names) = triple
```
ilan-gold, line 23: `_` for `dir_names` since it is not used?

mccalluc: I'll add an underscore: It might be useful in the future, and I'd never remember what's in that triple otherwise. Thanks!

PR #293, opened 2020-05-31
Mccalluc/stub description md

mccalluc:
I believe that Nils wants a description of each assay in markdown which we will render in the portal. (I have some concerns about how cleanly the assay types here will be translated into the portal: We'll make it work, but I think it's going to be more painful than you might expect.)

- @ngehlenborg : Can you clarify whether you're expecting the entire google doc for each assay to be translated, or a more brief synopsis? The audience for the documents we have is really the data submitters, and a document for data users might be different.
- @cebriggs7135 : Whatever Nils wants, are you able to commit to delivering it without assistance from me? There are plenty of resources on markdown, and with the other work on my plate, I would prefer not to be a copy-editor on this project.

If this is approved, I can move forward with integrating these stubs into the portal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.0.4 - In progress
 ### Added
+- Stub assay description MDs.
 - Generate unified description list.
 - Links to background docs.
 - Pre-fill enums in TSVs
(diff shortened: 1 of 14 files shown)


mccalluc:
Resolved by #294

PR #305, opened 2020-06-03
TO DO for bulkATACseq metadata

cebriggs7135:
This is a slash-delimit4d list of sequencing cycles-
name: sequencing_read_format
  description: Slash-delimited list of the number of sequencing cycles for, for example, Read1, i7 index, i5 index, and Read2.
  constraints:
    pattern: '\d+(/\d+)+'


mccalluc:
@cebriggs7135 : Is it always these four steps?

PR #311, opened 2020-06-03
TO DOs done for autofluorescence

cebriggs7135:
number_of_channels remains. Waiting for input from Jeff Spraggins.


cebriggs7135:
Input received. No further action needed other than separating level 2. All fields are currently under level 1 heading. This is inconsistent with the levels structure in all other assay schema.

PR #316, opened 2020-06-05
remaining TO DO for IMC: analyte_class / segment_data_format

cebriggs7135:
IMC: 
analyte_class


src/ingest_validation_tools/table-schemas/level-2/imc.yaml
```diff
@@ -15,7 +15,7 @@ fields:
   name: analyte_class
   constraints:
     enum:
-      - TODO
+      - imaging
```
cebriggs7135, line 18: proteins

src/ingest_validation_tools/table-schemas/level-2/imc.yaml
```diff
@@ -99,8 +99,7 @@ fields:
     enum:
       - um
 - name: segment_data_format
-  description: This refers to the data type, which is a float for the IMC counts.
-  type: number
+  description: This refers to the data type, which is a "Float" for the IMC counts.
```
ngehlenborg, line 102: @cebriggs7135 It looks like you removed the `type` field. It was `number` but now there is no longer a type for `segment_data_format`.

cebriggs7135: @ngehlenborg Yes because it actually is the word "Float", not a float number.

mccalluc: @cebriggs7135 :
- I'll make this an enum: `float`, `integer`, `string` should cover it.

src/ingest_validation_tools/table-schemas/level-2/imc.yaml
mccalluc: add `proteins` as an option? Will do.

PR #317, opened 2020-06-07
MxIF analyte_class edit

cebriggs7135:
proteins


src/ingest_validation_tools/table-schemas/level-2/mixif.yaml
```diff
@@ -14,7 +14,7 @@ fields:
   -
     name: analyte_class
     constraints:
-      required: False
+      required: proteins
```
ngehlenborg, line 17: @cebriggs7135 I believe that this field name is supposed to be "enum" rather than "required".

PR #339, opened 2020-07-07
Rename mixif.yaml to mxif.yaml

cebriggs7135:
The name of the tsv should be "mxif" rather that "mixif".


mccalluc:
@cebriggs7135 : Note that this will change the validation enum.

PR #348, opened 2020-08-11
Dir schema now allows descriptions

### LINKED ISSUE #346: Add a slot for description in directory schema
So when the files are in, we know what they mean.

mccalluc:
Fix #346.

Background: The first attempt at a "directory schema" serialized directories to a json structure, and then used json schema to validate. That didn't work well:
- Schemas hard to author.
- Schemas hard to read.
- Errors hard to interpret.

The second version gave up some flexibility, but is vastly improved in all three areas. Directory schemas are now nested yaml dicts, with globbing allowed on filenames where there will be variation.

Before this PR, at the leafs, we had either `0` or `1` for optional or required, but we want to record descriptions at least, and in the future indicate whether a particular file provides a QA report.

Questions:
- The code looks for `description` to know that a dict is a description of a file, and not a subdirectory. This would prevent `description` from being used as the name of a file. Ok?
- Optional files are indicated with `required: False`? It could just as easily be `optional: True`. Any preference?
- There are a lot of TODOs to fill in, and for many assay types we don't have schemas. @cebriggs7135 : Do you think it is feasible to get these filled in, or does it feel like we're digging ourselves into a hole?
- Do you think you'll want to support markdown in the descriptions? If so, I would like to change the field name to `description_md`.

### DIFF
--- a/src/ingest_validation_tools/directory-schemas/af.yaml
+++ b/src/ingest_validation_tools/directory-schemas/af.yaml
@@ -1,20 +1,20 @@
 LC Data:
-  '*': 0 # TODO: What is expected here?
+  '*': { required: False, description: "TODO" } # TODO: What is expected here?
 metadata:
   assay:
-    AF.csv: 1
+    AF.csv: { description: "TODO" }
 templates:
-  '*': 0 # TODO: What is expected here?
+  '*': { required: False, description: "TODO" } # TODO: What is expected here?
 VAN*-*K-*-*_*:
   CCF Metadata:
-    VAN*-*K-*-*_ccf_metadata: 1
+    VAN*-*K-*-*_ccf_metadata: { description: "TODO" }
   VAN*-*K-*-*_*-AF:
     processedMicroscopy:
       VAN*-*K-*-*-AF_preIMS_images:
-        VAN*-*K-*-*-AF_preIMS_registered.ome.tiff: 1
-      VAN*-*K-*-*-AF_preIMS_transformations: 1
-        transform_00_VAN*-*K-*-*-AF_preIMS_rigid.txt: 1
-        transform_01_VAN*-*K-*-*-AF_preIMS_affine.txt: 1
+        VAN*-*K-*-*-AF_preIMS_registered.ome.tiff: { description: "TODO" }
+      VAN*-*K-*-*-AF_preIMS_transformations: { description: "TODO" }
+        transform_00_VAN*-*K-*-*-AF_preIMS_rigid.txt: { description: "TODO" }
+        transform_01_VAN*-*K-*-*-AF_preIMS_affine.txt: { description: "TODO" }
       rawMicroscopy:
-        VAN*-*K-*-*-AF_preIMS_meta_unregistered.xml: 1
-        VAN*-*K-*-*-AF_preIMS_unregistered.czi: 1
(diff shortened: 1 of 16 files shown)


ngehlenborg:
Have you tried to make this compatible with the manifests that are being used for workflows? See here for an example: https://github.com/hubmapconsortium/sc-atac-seq-pipeline/blob/develop/create_snap_and_analyze-manifest.json

I would prefer a single format to define what should be contained in a dataset, what the files contain/represent (i.e., descriptions), potentially expected/delivered formats are, which files are QC reports, which files should be displayed in the UI vs only on Globus, etc.

src/ingest_validation_tools/directory-schemas/af.yaml
```diff
@@ -1,20 +1,20 @@
...
-  '*': 0 # TODO: What is expected here?
+  '*': { required: False, description: "TODO" } # TODO: What is expected here?
 metadata:
   assay:
-    AF.csv: 1
+    AF.csv: { description: "TODO" }
 templates:
-  '*': 0 # TODO: What is expected here?
+  '*': { required: False, description: "TODO" } # TODO: What is expected here?
 VAN*-*K-*-*_*:
```
ngehlenborg, line 8: As defined here, AF would only support *K*idney data from *VAN*derbilt. That is too restrictive.

mccalluc: I've made an issue for Chris: #349

mccalluc:
> I would prefer a single format to define what should be contained in a dataset, what the files contain/represent (i.e., descriptions), potentially expected/delivered formats are, which files are QC reports, which files should be displayed in the UI vs only on Globus, etc.

I think we could use the same keys in each to represent this information as it becomes available, but I would be reluctant to change the overall structure to match:
- Non-devs (and even younger developers) have lot of trouble using regexes successfully. I think globs are sufficient for this job, and can be used correctly by curators, and read by TMCs.
- The nested structure isn't essential, but I think it makes it much easier for TMCs to read, and it avoids the kinds of path inconsistencies which will be inevitable if the whole path needs to be repeated for every leaf in a deeply nested structure.

mccalluc:
From Nils:
>  I think we need to have a conversation with the TMCs about the directory validation as part of the planning for the upcoming data submissions. I am a bit concerned that we make too many assumptions about their process.
>
> I think @Chris B should comment on the PR, but I don’t think we can finalize this without TMC input. I anticipate that the next round of submissions might look different from the first.

I'll change this to a draft PR until we have the necessary information.

cebriggs7135:
@mccalluc 
A. I could understand better if you could show me examples.

B. Answers to your questions:

The code looks for description to know that a dict is a description of a file, and not a subdirectory. This would prevent description from being used as the name of a file. Ok?
CB: Sounds right.

Optional files are indicated with required: False? It could just as easily be optional: True. Any preference?
CB: False is easy to understand.

There are a lot of TODOs to fill in, and for many assay types we don't have schemas. @cebriggs7135 : Do you think it is feasible to get these filled in, or does it feel like we're digging ourselves into a hole?
CB: Let me use that script you gave me awhile ago to extract all the TODOs so I can take a look and answer your question.

Do you think you'll want to support markdown in the descriptions? If so, I would like to change the field name to description_md.
CB: I have no preference unless doing something other than MD requires I learn something new.

PR #355, opened 2020-08-20
Mccalluc/dir validation as array of dicts

mccalluc:
- @ngehlenborg : If you have any comments about how the directory schema is shown in the generated docs? 
- @cebriggs7135 : Do the directory schemas and the errors make sense? Will you be able to fill in new schemas as you learn what the TMCs will be delivering? Will you be able to fill out all the schemas that are just stubbed?
- @mruffalo : Nils wants to use the same structure at ingest time as you use to filter the pipeline outputs. The semantics aren't quite the same (you are filtering, rather than validating, so you don't care whether a given file is required or optional), but otherwise what we're doing is fairly analogous. The hope is that as we add features (for example, flagging files) we'll be able to use the same tags in both environments. Does this sound ok?


ngehlenborg (commented):
Please remove tables from docs for now.

mccalluc:
Take out the generetad section from docs.

mccalluc:
Removing directory structure tables from docs; will merge when tests pass.

PR #360, opened 2020-09-21
Mccalluc/arbitrary validation code

### LINKED ISSUE #361: Deliver hooks for deeper validation to Joel (IEC)
We want to be able to run code that can look inside files and make sure they make sense for the pipeline.

Should the hook code be:
- Python functions?
- Arbitrary executables? (and hope the user has the right dependencies in place)
- Docker images?
- CWL?

mccalluc:
This gives Joel the ability to add code that can validate the content of a submission. (I wasn't able to select Joel as a reviewer, so there might be something odd about the permissions on the this repo...). Branching from the other PR to lessen the risk of merge conflicts.

The return values from `collect_errors` are just serialized as YAML for display: I've been using very human-readable keys so the error report makes sense, even if you don't know YAML.

Fix #361


jswelling (commented):
I see how to add assay-specific routines within src/ingest_validation_tools/content_validation.  It would be helpful to have a hook for generic tests that must be applied to all dataset submissions, for example the consistency of provenance of the donor and sample.  I think we will also want to apply some generic tests by file type, for example checking the internal validity of OME-TIFF files.  One possibility that might make an easy interface would be to have directory_validator:validate_directory_content() call a factory function passing the path and assay type, and then call collect_errors() on the returned object.  We could then provide the factory function.

mccalluc:
There are several different requests here, and I would like to handle them separately. If the functionality in this PR is not useful, let me know and it can be closed.

New requests:
- "a hook for generic tests that must be applied to all dataset submissions, for example the consistency of provenance of the donor and sample"
- "generic tests by file type, for example checking the internal validity of OME-TIFF files"
- "One possibility that might make an easy interface would be to have directory_validator:validate_directory_content() call a factory function passing the path and assay type, and then call collect_errors() on the returned object. We could then provide the factory function."

jswelling:
The third ask, that a factory function be used, subsumes the first two requests because we could then easily implement those two features.

mccalluc:
Tests are failing because of "sn" vs "sc" mismatch in terms, which I think is because this branch and master are pointing at different versions of `search-api`: I've asked Zhou if we can try to get that master up to date soon.

mccalluc:
@jswelling : Is this a shape you can work with? I think [this](https://github.com/hubmapconsortium/ingest-validation-tools/pull/360/files#diff-e939a445939b94f41ba36f7e523a5c9aR5-R8) is the point you'll be extending from: Let me know if you don't think that will work.

jswelling (commented):
So how do I add additional tests from outside the module?  My plan was to inject them via the globally visible DIRECTORY_CONTENT_VALIDATOR_FACTORY, as in:

```
import ingest_validation_tools
ingest_validation_tools.directory_validator.DIRECTORY_CONTENT_VALIDATOR_FACTORY=...my version...
```

Also, if the validator returns a dict, what's the intention for the meaning of the keys?

This would allow me to add ingest-time checks that would be impractical to run locally.

mccalluc:
> So how do I add additional tests from outside the module?

If you don't think you'll want to add code in this repo, I would really prefer that we set a new repo that can be a dependency of `ingest-validation-tools`, rather than making `ingest-validation-tools` a dependency of something else. There can be flags to skip tests that are very long-running, or that require resources behind the firewall, or that have hard-to-install dependencies, but I really think we'll be better off if we don't start from an architecture that makes it impossible for the TMCs to run the same checks we'll be running.

> This would allow me to add ingest-time checks that would be impractical to run locally.

I think there's a whole lot of low-hanging fruit we can handle before that becomes an issue. I'd prefer getting in more checks that can run locally before getting to more esoteric checks.

> Also, if the validator returns a dict, what's the intention for the meaning of the keys?

All the keys are human readable, and for now, the whole thing is just rendered as YAML. [For example](https://github.com/hubmapconsortium/ingest-validation-tools/tree/master/dataset-examples/bad-scatacseq-data). The keys are just headings on the errors described in the value: It might be a file path, or a line within a file, or a grouping of errors.

If this isn't persuasive, it may just be easier to run your checks entirely separately from what's being done here. The only thing that's gained is having the single output that combines all the errors, but that's pretty minor.

jswelling:
To be explicit, I think a validation process like this clearly calls for a pluggable architecture.  I think the tests themselves should be separable from the mechanism which applies the tests, and that an external caller should be able to add tests.  There are a number of patterns that would allow this.  The factory function approach we've discussed just seemed like the most convenient given the existing code.  I gather that your model is that any program invoking the validation process would do so by running it from the command line, but I'd argue that the necessity of a command line invocation is not very Pythonic.
I agree that it is desirable for a user at a TMC to be able to run all tests locally, but I think the most common use case will involve testing some parts of a data submission locally (like the metadata.tsv file), and assembling the bulk of the submission in place on Globus.  I think that by far the most common use case will involve turning off some of the tests.  We plan to provide quick feedback when the 'validate' button is pressed in the UI, and we'll need to divide tests into fast and slow to do that.
With respect to displaying the errors as YAML, I find that a bit kludgy. ingest-ui will also need to display the results, and may provide more formatting.  I'd like the return type to be explicit, with a schema if necessary.

mccalluc:
@jswelling -- With the new PR you've started, should this be closed?

mccalluc:
This has been superseded.
> I think it can be closed- we've moved on.  Next steps would be (1) I put a meaningful test in ingest-validation-tests and present a PR whereby it gets called by ingest-validation tools, and (2) we work out some kind of mechanism where some or all the plug-in tests can not be called.

PR #370, opened 2020-09-28
scatac dir schema

### LINKED ISSUE #451: Directory schema for scatacseq
Just `.*\.fastq\.gz` with no description right now.

mccalluc:
Fix #451 

For many of the sub-types here, I don't have access to any Globus examples, so I'm not at all confident that this is an accurate descriptions of the correct structure.


mccalluc:
Closing PR to clarify that I do not have a role in authoring directory schemas... but I'll keep the branch around, in case it's useful.

PR #372, opened 2020-09-30
Welling/arbitrary validation code

jswelling:
Modified to provide a hook for easier addition of functionality from outside this repo.

### DIFF
--- a/src/ingest_validation_tools/directory_validator.py
+++ b/src/ingest_validation_tools/directory_validator.py
@@ -3,6 +3,7 @@ import re
 from fnmatch import fnmatch
 from importlib import import_module
 
+
 def default_factory(base_path, assay_type):
     """
     default validator factory
@@ -30,6 +31,7 @@ class DefaultDirectoryContentValidator(DirectoryContentValidator):
     """
     Implements backward compatible calls to content validation modules
     """
+
     def collect_errors(self):
         validator = import_module(
             f'ingest_validation_tools.content_validation.{self.assay_type}')
--- a/src/ingest_validation_tools/directory_validator.py
+++ b/src/ingest_validation_tools/directory_validator.py
@@ -3,6 +3,13 @@ import re
 from fnmatch import fnmatch
 from importlib import import_module
 
+def default_factory(base_path, assay_type):
+    """
+    default validator factory
+    """
+    return DefaultDirectoryContentValidator(base_path, assay_type)
+
+
 DIRECTORY_CONTENT_VALIDATOR_FACTORY = default_factory
 
 
@@ -27,13 +34,6 @@ class DefaultDirectoryContentValidator(DirectoryContentValidator):
         validator = import_module(
             f'ingest_validation_tools.content_validation.{self.assay_type}')
         return validator.collect_errors(self.base_path)
-
-
-def default_factory(base_path, assay_type):
-    """
-    default validator factory
-    """
-    return DefaultDirectoryContentValidator(base_path, assay_type)
 
 
 class DirectoryValidationErrors(Exception):
--- a/src/ingest_validation_tools/directory_validator.py
+++ b/src/ingest_validation_tools/directory_validator.py
@@ -3,6 +3,38 @@ import re
 from fnmatch import fnmatch
 from importlib import import_module
 
+DIRECTORY_CONTENT_VALIDATOR_FACTORY = default_factory
+
+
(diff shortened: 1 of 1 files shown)


src/ingest_validation_tools/directory_validator.py
```diff
@@ -3,6 +3,38 @@
...
+
+
+class DirectoryContentValidator(object):
+    def __init__(self, base_path, assay_type):
+        self.base_path = base_path
+        self.assay_type = assay_type
+
+    def collect_errors(self):
+        """
+        returns string or None.  Multiple errors are to be reported
```
mccalluc, line 23: Is there are reason not to return an array? For that matter, I'm not sure restricting to strings is necessary, if dicts could be useful.

src/ingest_validation_tools/directory_validator.py
```diff
@@ -3,6 +3,38 @@
...
+        via a single multi-line string.
+        """
+        return None
+
+
+class DefaultDirectoryContentValidator(DirectoryContentValidator):
+    """
+    Implements backward compatible calls to content validation modules
```
mccalluc, line 31: This is backward compatibility to the stubs I had created? Unless you see you see yourself using them, they don't need to (and shouldn't) stay around.

mccalluc (commented):
Couple comments: I'll fix the linting locally, and see if there other issues. (Are you able to run test.sh locally? - It's fine with me to push commits without running tests each time, and use github as a backup, but I want to make sure everything works for you.)

My main question is whether it's ok to get rid of `af.py` etc. If you don't need them, I'm not attached to them.

src/ingest_validation_tools/directory_validator.py
```diff
@@ -4,6 +4,40 @@
...
+def default_factory(base_path, assay_type):
+    """
+    default validator factory
+    """
+    return DefaultDirectoryContentValidator(base_path, assay_type)
+
+
+DIRECTORY_CONTENT_VALIDATOR_FACTORY = default_factory
+
+
```
mccalluc, line 16:
```suggestion
def directory_content_validator_factory(base_path, assay_type):
    """
    default validator factory
    """
    return DefaultDirectoryContentValidator(base_path, assay_type)


```
... and the corresponding change below, if you feel like it? I guess the original name makes me think you would add more factories over time... but I think the advantage of the Factory pattern would be hiding the details inside here, instead.

mccalluc (commented):
I fixed the whitespace issue and committed to this branch... Hope that's ok?

Could you add one `DirectoryContentValidator` subclass for one type so I can see how this is going to grow? (A no-op stub would be fine.)

src/ingest_validation_tools/directory_validator.py
jswelling: Sorry, I thought string or none was your intention with the design.  I'd prefer returning an array, as long as we specify the types of the array elements.

src/ingest_validation_tools/directory_validator.py
jswelling: Yeah, that's backward compatibility.  I don't object to removal, but if the factory plug-in has not been externally set it still has to work sensibly.

jswelling: wrt lint and formatting, I'm not using a proper IDE for these edits so I apologize for breaking PEP8 & etc.  Thanks for fixing things up.  I was running test_dataset_examples.sh but I'm not set up to run travis locally.

src/ingest_validation_tools/directory_validator.py
jswelling: Agreed on both counts.  The problem with providing an example is that I plan to do it in the context of the larger piece of code which imports ingest-validation-tools, but I could add an explicit example.  I tested it by modifying ingest_validation_tools/validation_utils to overlay a substitute factory method at the time of import.  I could add an example to demonstrate that overlay but it'll maybe be more complex than one wants an example to be.  I thought the pattern was common enough that a specific example wouldn't be needed.

mccalluc:
@jswelling : If it's ok, I'll merge these changes to my branch, and then do some more work and ask you to take a final look at it, before merging to master. Thanks for helping with this.

jswelling:
Yes, that's OK by me.  I'll test replacing the callback in the context of ingest-pipeline once it has been merged to your branch.

PR #374, opened 2020-10-05
Directory schema for nano protocols

### LINKED ISSUE #450: Directory schema for nano-*

mccalluc:
Fix #450 

Copying comments from Slack:
> - nanoPOTs and nanoDESI are new … Are they one type? nanopots-nanodesi ?
> - Does each dataset contain multiple regions?
> - Folder name (and others) has a space in it …_Imaging region001_… … Do we want to push back / clarify this?
> - Is the metadata.csv our metadata? or something else?
> - Experiment description needs to be on protocols.io: I think we’re prefer not to repeat that here.
> - Powerpoint is really not a reusable format… appropriate to push back?
> - The PNGs aren’t doing any harm… but I think we should be steering the labs towards just providing the data.

and beyond all of that, I think we should also hold ourselves to filling in descriptions as they come in: No time like the present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - ng/ul to nM.
 - Change to flat directory schema structure.
 - Update README.
+- Directory Schema for nano protocols.
 
 ## v0.0.4 - 2020-06-26
 ### Added
--- a/README-generate_dataset_docs.py.md
+++ b/README-generate_dataset_docs.py.md
@@ -1,10 +1,10 @@
 ```
 usage: generate_dataset_docs.py [-h]
-                                {af,bulkatacseq,bulkrnaseq,codex,imc,lcms,maldiims,mxif,scatacseq,scrnaseq,seqfish,stained,wgs}
+                                {af,bulkatacseq,bulkrnaseq,codex,imc,lcms,maldiims,mxif,nano,scatacseq,scrnaseq,seqfish,stained,wgs}
                                 target
 
 positional arguments:
-  {af,bulkatacseq,bulkrnaseq,codex,imc,lcms,maldiims,mxif,scatacseq,scrnaseq,seqfish,stained,wgs}
+  {af,bulkatacseq,bulkrnaseq,codex,imc,lcms,maldiims,mxif,nano,scatacseq,scrnaseq,seqfish,stained,wgs}
                         What type to generate
   target                Directory to write output to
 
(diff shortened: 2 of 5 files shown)


cebriggs7135:
Here are details on the folder content from Manxi:
1.	Field content:
•	“experiment.docx” will contain the experimental conditions, instrument settings, experimental goal, etc. 
My comment: _All experimental protocols should be registered in protocols.io and the link will go into the protocols.io fields in the  assay_metadata.tsv you refer to below>>_
•	“metadata.csv” is the info on that excel template you sent me. 
•	“sample info.pptx” is several slides containing the sample preparation, where the sample come from,  scanning info and autoflurescence image of the scanning region for comparison. 
My comment_Rather than a pptx, we may ask for that as a pdf – to be determined._
•	“raw data.imzml” file is a imzML file generated from the raw data set. 
•	“peak assignment.csv” is an excel file contains the peaks with its assigned identity. 
•	“ion images” contains a set of images of lipids/metabolites with an interesting distribution.
2.	Each imaging region folder contains the same collection of files.
3.	“section” is tissue, I think for each tissue, we are just gonna collect one dataset.
4.	Each m/z png file is a picture of the ion images.

cebriggs7135:
- nanoPOTs and nanoDESI are new … Are they one type? nanopots-nanodesi ?
They both are spatial mass spec. assays that synthesize images from the mass spec. data. nanoPOTS starts with liquid chromatography of nanodroplets of sample to separate proteins before the mass spec. analysis.
- Does each dataset contain multiple regions?
My assumption is that each dataset represents one tissue sample with data collected on several regions of that tissue sample. Data on each region is provided in a folder. The directory contains one folder for each region of the tissue sample.
- Folder name (and others) has a space in it …Imaging region001… … Do we want to push back / clarify this?
- Is the metadata.csv our metadata? or something else?
- Experiment description needs to be on protocols.io: I think we’re prefer not to repeat that here.
- Powerpoint is really not a reusable format… appropriate to push back?
It may be the best way to provide data or we could require they be converted to pdfs.
- The PNGs aren’t doing any harm… but I think we should be steering the labs towards just providing the data.

mccalluc:
Chris clarifies that Chuck has no responsibilities at this moment.

cebriggs7135:
UPDATE
Manxi emailed the following regarding the possibility of having a single assay metadata form for both nanoDESI & nanoPOTS AND the possibility of a few DESI vs POTS fields:
_I assume we will have different added fields in metadata, but I will discuss with them this Friday in our meeting. Yes, sure, I will send you the spreadsheet once it’s done._

cebriggs7135:
Update:
Screen Shot 2020-10-13 at 12 24 52 PM

mccalluc:
Closing PR to clarify that I do not have a role in authoring directory schemas... but I'll keep the branch around, in case it's useful.

PR #378, opened 2020-10-13
Add "extras" to every dir schema

mccalluc:
Nils, as discussed, each dataset directory could have a conventional sub-directory that is loose. Before merging, I would move this into the code rather than having this copy and paste... but I think this form makes the intention clear.

Are the name of the directory and the description appropriate?


ngehlenborg (approved):

ropelews (approved):
No objection, but might want to clarify for TMCs -- 1) Any files/directories in this extras directory will not be validated 2) TMCs should not use this directory for higher level data submission (i.e. tmc processed data) that data should go through a separate ingestion process.

PR #379, opened 2020-10-13
Mccalluc/prettier html

mccalluc:
Improves the HTML rendering of errors. This is available from the commandline: The browser might still be a better place to look at errors, even if it's being run locally: I'd suggest that the web-app on your side also call `as_html_fragment`, rather than duplicating efforts.
```
src/validate_submission.py --local_directory dataset-examples/bad-mixed/submission/ --output as_browser
```
produces:
Screen Shot 2020-10-13 at 2 59 19 PM

- Bill: Does the `as_html_fragment` here look like something you'll be able to use?
- Joel: Does this rendering allay your concerns about the lack of a tight schema for the output?


jswelling:
Whatever ingest-validation-tools returns, I need to pass to ingest-api within an existing JSON structure. Historically the error info was displayed raw in a pop-up, but that feature is disabled now pending some restructuring of the 'Submit'/'Invalid' process.  I don't object to passing the html back in a string if that format will work for Bill's people.  I presume "no errors" would still be signaled by an empty string.

WRT the structuring of the reply, I'm satisfied with shifting that question off to the structure of the return value of Validator.get_errors(), which I think we've agreed on.

mccalluc:
> I presume "no errors" would still be signaled by an empty string.

Good catch: The behavior had been:
```
src/validate_submission.py --type_metadata codex   dataset-examples/good-codex-akoya/submission/codex-akoya-metadata.tsv --output as_html
<html><body><pre>No errors!
</pre></body></html>
```
and I need to restore that in this branch... For the use case where it's being run on the CLI, but the browser is being used for display, there does need to be an explicit `No errors!` message.

I agree that passing html through and injecting it is not what normal separation-of-concerns would dictate, but since the same rendering functionality is needed on both ends, I'd suggest that the work not be duplicated.

> WRT the structuring of the reply, I'm satisfied with shifting that question off to the structure of the return value of Validator.get_errors(), which I think we've agreed on.

Those lists would be one element in the larger structure, one example of which is in the screen shot.

If that's not the result you want, suggest migrating to slack or in-person, rather than continuing thread here. Two possible claims, and you might have others in mind:
- Flatter structure necessary for rendering: No, this PR rendering arbitrary nestings.
- Flatter structure is more usable: Perhaps... but I'm looking at the example above, and in the examples, and while the messages could be flattened to two strings, one to label the list, and one for the items in the list, I really don't think that would be easier to understand and work with.

jswelling:
See also https://github.com/hubmapconsortium/ingest-validation-tools/issues/388

ingest-pipeline can cope with varying details here more easily than ingest-ui, so @shirey 's input will be crucial here.

shirey:
@mccalluc To echo the message that I just left in #338 I'd prefer just a list of error messages as string in non-techy-human-readable fashion so I can format and display in the ingest-ui.  I don't want a "No errors!" message, but would prefer a flag from Joel telling me if there are errors or not (an empty list would work too to designate no errors, but a flag would be nice).

mccalluc:
It's coming in a new PR. This form is more readable, and will be available in the CLI, even if not in the ingest-ui.

PR #380, opened 2020-10-14
Table schema for contributors.tsv

mccalluc:
Lots of little code details and updates to the fixtures, but the point I'd like confirmation on is the
[table schema for contributors.tsv](https://github.com/hubmapconsortium/ingest-validation-tools/compare/mccalluc/add-contributors...mccalluc/add-contributors-continued?expand=1#diff-14542c76726df7e6c04aac84ca44dd835d4783f02c7cb8da8d514249083b51e4).

Proposal is to have two columns: The ORCID ID, and optionally, a human-readable name, with the idea that if the name doesn't match what the ORCID API returns, it will be a warning. @shirey : We've wondered if you need any kind of role field? If not, great! Or if so, that can be accommodated too.


dataset-examples/bad-codex-data/submission/contributors.tsv
```diff
@@ -0,0 +1 @@
+orcid_id	name
```
ropelews, line 1: Having gone through doi issuance for the Brain Image Library project, the attached screen capture is how we collect the data for doi issuance by DataCite. We use the table to collect both creator and contributor information, having a field asking if the contributor is also a creator. A subset of the relevant DataCite tags for contributorType are also listed.  Contributors can be institutions or research groups which typiccally do not have orcid's but rather RORs or RRIDs.
Contributors

ropelews (commented):
See comment on contributors.tsv and evaluate expansion of the table to collect additional information shown for completion of a DataCite doi contributors (and creators) tables.

mccalluc:
[In Slack](https://hubmapconsortium.slack.com/archives/C01BKNPURDE/p1603304869096800) we concluded that a smaller set of fields would suffice:
> Ok, Bill, I’ll collect the 6 fields in your schema (affiliation, first_name, last_name, middle_name_or_initial, name, orcid_id) and warn if the names supplied do not match what ORCID has on file. This can change over time, but I think for now matching the existing database makes sense.

mccalluc:
travis-ci.org still stalled. Other tests pass.

PR #382, opened 2020-10-15
Add base classes for plugin Validator and ValidatorError

jswelling:
Beginning integration of plugin validators, as discussed.  This provides the base class but not yet the mechanism by which plugins are applied.

### DIFF
--- a/src/ingest_validation_tools/plugin_validator.py
+++ b/src/ingest_validation_tools/plugin_validator.py
@@ -47,6 +47,16 @@ class Validator(object):
 
 def run_plugin_validators_iter(metadata_path: PathOrStr,
                                plugin_dir: PathOrStr) -> Iterator[KeyValuePair]:
+    """
+    Given a metadata.tsv file and a path to a directory of Validator plugins, iterate through the
+    results of applying each plugin to each row of the metadata.tsv file.  The 'assay_type' field
+    of the metadata.tsv file is used to provide the assay, rather than the prefix of the filename.
+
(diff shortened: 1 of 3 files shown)


src/ingest_validation_tools/plugin_validator.py
```diff
@@ -0,0 +1,45 @@
...
+    def collect_errors(self) -> List[str]:
+        """
+        Returns a list of strings, each of which is a
+        human-readable error message.
+
+        "No error" is represented by returning an empty list.
+        If the assay_type is not one for which this validator is intended,
+        just return an empty list.
+        """
+        return []
```
mccalluc, line 45:
```suggestion
        throw NotImplementedError()
```

mccalluc (approved):
I should read up on typing. Perhaps throw an exception if someone tries to run a method from the abstract class? What do you think about linting going forward? Maybe you don't see yourself contributing that much to this repo going forward, but if you do, would pre-commit-hooks or black formatting make your life easier?

I'm comfortable with PR authors merging once approved. After that, is the next step in your court, or should I move this forward?

jswelling:
There are two next steps; starting to populate the repo of plug-in tests and calling the validators from ingest-validation-tools.  I can do a PR against ingest-validation-tests to call the plugins if you'd like.  Could you place the submodule for ingest-validation-tests where you want it within ingest-validation-tools?  

I'm learning Python type hinting from reading Matt's commits.  If nothing else I think it improves the clarity of the descriptions, but I'm not very good at it yet.

PR #387, opened 2020-10-19
ingest-validation-tests submodule

mccalluc:
I'm really fine with putting this wherever ... move down to src if you'd prefer? It can also be moved later.

When it's actually being used we'll need to update the README, but not needed yet.

I've had trouble moving the directory validation work forward, and that will be blocking you at some point. Right now nearly every assay just is expecting `TODO.txt`... I'd like to get it filled in with the right think, but as a stop gap, those could all be changed to `.+`?


jswelling:
If it were moved down to src/ingest_validation_tools it would be possible to import it with a relative import, like 'import ..ingest_validation_tests' .  It'll still have to be overridden sometimes, but I'd think that might reduce the need for unpleasant path logic.

mccalluc:
@jswelling : I've moved the submodule down to src/ingest_validation_tools/ingest-validation-tests... but I'm thinking the dashes will cause problems when trying to import? Anyway: if you'd like other changes in this PR, please let me know, but I would be fine with going with this now, and sorting out the import problems when there's actually code to run.

mccalluc:
@jswelling : If this is still useful, approve or let me know what should change. If it's been handled elsewhere, can be closed.

mccalluc:
> No worries! I guess the submodule path of ingest-validation-tests is still an open question, so in that respect #387 is still relevant.

Keeping open.

mccalluc:
Merging this, and will follow up with #392.

PR #389, opened 2020-10-19
Mccalluc/flat option

mccalluc:
```
$ src/validate_submission.py --local_directory dataset-examples/bad-mixed/submission/ --output as_text_list
Metadata TSV Errors: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (as codex-akoya): Internal: The value "-INVALID-" in row 2 and column 1 ("A") does not conform to the pattern constraint of "[A-Z]+[0-9]+"
Metadata TSV Errors: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (as codex-akoya): External: codex-akoya-metadata.tsv (row 2): Not allowed: not-good-for-either-type.txt
Metadata TSV Errors: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (as codex-akoya): External: codex-akoya-metadata.tsv (row 2): Required but missing: channelnames\.txt
Metadata TSV Errors: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (as codex-akoya): External: codex-akoya-metadata.tsv (row 2): Required but missing: experiment\.json
Metadata TSV Errors: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (as codex-akoya): External: codex-akoya-metadata.tsv (row 2): Required but missing: exposure_times\.txt
Metadata TSV Errors: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (as codex-akoya): External: codex-akoya-metadata.tsv (row 2): Required but missing: cyc.*_reg.*_.*/.*_.*_Z.*_CH.*
Metadata TSV Errors: dataset-examples/bad-mixed/submission/scatacseq-metadata.tsv (as scatacseq): Internal: The value "-INVALID-" in row 2 and column 1 ("A") does not conform to the pattern constraint of "[A-Z]+[0-9]+"
Metadata TSV Errors: dataset-examples/bad-mixed/submission/scatacseq-metadata.tsv (as scatacseq): Internal: Column 17 ("Q") is a required field, but row 2 has no value
Metadata TSV Errors: dataset-examples/bad-mixed/submission/scatacseq-metadata.tsv (as scatacseq): Internal: Column 27 ("AA") is a required field, but row 2 has no value
Metadata TSV Errors: dataset-examples/bad-mixed/submission/scatacseq-metadata.tsv (as scatacseq): External: scatacseq-metadata.tsv (row 2): Not allowed: not-good-for-either-type.txt
Metadata TSV Errors: dataset-examples/bad-mixed/submission/scatacseq-metadata.tsv (as scatacseq): External: scatacseq-metadata.tsv (row 2): Required but missing: .*\.fastq\.gz
Reference Errors: Multiple References: bad-shared-dataset: dataset-examples/bad-mixed/submission/codex-akoya-metadata.tsv (row 2)
Reference Errors: Multiple References: bad-shared-dataset: dataset-examples/bad-mixed/submission/scatacseq-metadata.tsv (row 2)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Update README, and diagram.
 - Prettier HTML output.
 - Add donor.yaml, where we can explain donor metadata fields, and hook it into field-descriptions.yaml.
+- Add as_text_list option.
 
 ## v0.0.4 - 2020-06-26
 ### Added
--- a/README-validate_samples.py.md
+++ b/README-validate_samples.py.md
@@ -1,11 +1,11 @@
 ```
 usage: validate_samples.py [-h] --path PATH
-                           [--output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_yaml}]
+                           [--output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_text_list,as_yaml}]
 
 Validate a HuBMAP Sample metadata TSV.
 
 optional arguments:
   -h, --help            show this help message and exit
   --path PATH           Sample metadata.tsv path.
-  --output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_yaml}
+  --output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_text_list,as_yaml}
 ```
--- a/README-validate_submission.py.md
+++ b/README-validate_submission.py.md
@@ -5,7 +5,7 @@ usage: validate_submission.py [-h]
                               [--optional_fields FIELD [FIELD ...]]
                               [--dataset_ignore_globs GLOB [GLOB ...]]
                               [--submission_ignore_globs GLOB [GLOB ...]]
-                              [--output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_yaml}]
+                              [--output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_text_list,as_yaml}]
                               [--add_notes]
 
 Validate a HuBMAP submission, both the metadata TSVs, and the datasets,
@@ -34,7 +34,7 @@ optional arguments:
   --submission_ignore_globs GLOB [GLOB ...]
                         Matching sub-directories in the submission will be
                         ignored.
-  --output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_yaml}
+  --output {as_browser,as_html_doc,as_html_fragment,as_md,as_text,as_text_list,as_yaml}
   --add_notes           Append a context note to error reports.
 
 Typical usecases:
--- a/src/ingest_validation_tools/error_report.py
+++ b/src/ingest_validation_tools/error_report.py
@@ -4,6 +4,8 @@ from webbrowser import open_new_tab
 from pathlib import Path
 from yattag import Doc, indent
 
+from ingest_validation_tools.message_munger import munge
+
 
 # Force dump not to use alias syntax.
 # https://stackoverflow.com/questions/13518819/avoid-references-in-pyyaml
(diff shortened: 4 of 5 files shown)


jswelling:
So this is basically a csv file with ':' as a separator.  A tab-separated form might be safer, in that there would be no concern about ':' in the error messages produced by plug-ins.

mccalluc:
> So this is basically a csv file with ':' as a separator. 

I wouldn't look at it that way. Bill wants a flat list of human readable strings:

> Can we get a simple list of human (non-programmer) readable error messages- one string per found problem?

and colons are pretty human readable. It could be done other ways... the order of each line could be reversed, the colon could be replaced with "->" or "👉". 

If you want something other than a flat list of human-readable strings, or you don't think these are human-readable, let me know.

jswelling:
@mccalluc I think @shirey 's actual need is to get an ordered list of error messages which ingest-ui can then format as it sees fit.  I propose that the returned lines of text literally constitute a .tsv .  (no errors means no lines of text).  ingest-pipeline can catch that .tsv, parse it as it usually parses .tsv data, and present the results to Bill's team either in tabular form or as a JSON data structure.  This would allow us to use existing tsv-parsing tools to roll up the error  lines into whatever form Bill's team finds most convenient.

shirey:
@mccalluc Just to confirm, when calling from within Python I'll get back a Python List of strings, one error message per string?  I'd prefer better, non-techy (without Regex expressions) and chopped fields. Just a single, correctly formed sentence (or two)-  Something, for example, like "In row 2, a file containing experimental data named experiment.json is required, but was not found."  or "Column 17, row 2 should contain a value for XXXXX, but nothing was found."  I can use this for now, but it will be confusing to many of the users.

mccalluc:
@shirey :
> when calling from within Python I'll get back a Python List of strings, one error message per string

Yes.

> non-techy (without Regex expressions)

Some of the messages are coming from the table validator package. I believe that a message that tells you what's wrong is better than one that just tells you something is wrong.

I'll add some regexes to munge the output. Thank you for clarifying what you want.

mccalluc:
```
$ src/validate_submission.py --local_directory dataset-examples/bad-mixed/submission/ --output as_text_list
In codex-akoya-metadata.tsv, the value "-INVALID-" in row 2 and column 1 ("A") is not the right form.
In the dataset referenced by codex-akoya-metadata.tsv (row 2), the file "not-good-for-either-type.txt" is not allowed.
In the dataset referenced by codex-akoya-metadata.tsv (row 2), a file matching "channelnames\.txt" is required but missing.
In the dataset referenced by codex-akoya-metadata.tsv (row 2), a file matching "experiment\.json" is required but missing.
In the dataset referenced by codex-akoya-metadata.tsv (row 2), a file matching "exposure_times\.txt" is required but missing.
In the dataset referenced by codex-akoya-metadata.tsv (row 2), a file matching "cyc.*_reg.*_.*/.*_.*_Z.*_CH.*" is required but missing.
In scatacseq-metadata.tsv, the value "-INVALID-" in row 2 and column 1 ("A") is not the right form.
In scatacseq-metadata.tsv, column 17 ("Q") is a required field, but row 2 has no value.
In scatacseq-metadata.tsv, column 27 ("AA") is a required field, but row 2 has no value.
In the dataset referenced by scatacseq-metadata.tsv (row 2), the file "not-good-for-either-type.txt" is not allowed.
In the dataset referenced by scatacseq-metadata.tsv (row 2), a file matching ".*\.fastq\.gz" is required but missing.
The dataset directory "bad-shared-dataset" is referenced by codex-akoya-metadata.tsv (row 2), and somewhere else, but references should be unique.
The dataset directory "bad-shared-dataset" is referenced by scatacseq-metadata.tsv (row 2), and somewhere else, but references should be unique.
```

mccalluc:
Travis.org still stalled. Travis.com is happy. Merging.

PR #391, opened 2020-10-23
Revert "Add base classes for plugin Validator and ValidatorError"

jswelling:
Reverts hubmapconsortium/ingest-validation-tools#382 ,  I've restored the welling/plugin_tests branch, so once this is reverted we can create a new PR for review.


mccalluc (commented):
From slack: 
> .... I think it’d be simplest to get what’s there now working (it looks good!), but if you’d really prefer to revert, I won’t block that: Thank you for being careful!

So your call: Thanks for setting up this PR, but I think getting the tests passing and making a couple other tweaks doesn't require a full revert.

mccalluc:
Decided just to move forward.

PR #392, opened 2020-10-23
Welling/plugin tests 2

jswelling:
This merge contains PEP8 fixes and response to comments wrt #382.  Remaining issues:

* The correct value of PLUGIN_DIR_REL_PATH at submission.py:16 has not yet been set.
* Perhaps the ValidationErrors should be explicitly caught at some level and converted to text error messages.
* Arguably plugin_validator.py should actually be in the ingest_validation_tests repo rather than here, because the Validator class is only used in that repo.


src/ingest_validation_tools/plugin_validator.py
```diff
@@ -71,31 +72,35 @@ def run_plugin_validators_iter(metadata_path: PathOrStr,
...
 
+    if all('data_path' in row for row in rows):
+        for row in rows:
+            data_path = Path(row['data_path'])
+            if not data_path.is_absolute():
+                data_path = (metadata_path / data_path).resolve()
+            for k, v in validation_error_iter(data_path, assay_type, plugin_dir):
+                yield k, v
+    else:
+        raise ValidatorError(f'{metadata_path} has no "data_path" column')
```
mccalluc, line 88: Preference to rely on these checks being done at a higher level, and only run these checks when the higher level checks pass, and not to add exceptions we don't recover from here. If we did hit this, it would be a bug, and I'm not sure that these `raises` make it easier to debug.

Let's not belabor this now, but where I'm coming from might make more sense when you get to the `bad` test fixtures.

src/ingest_validation_tools/submission.py
```diff
@@ -106,7 +107,7 @@ def _get_single_tsv_external_errors(self, type, path):
         errors = {}
         rows = _get_tsv_rows(path)
         if not rows:
-            errors['Warning'] = f'File has no data rows.'
+            errors['Warning'] = f'File {Path(path).name} has no data rows.'
```
mccalluc, line 110:
```suggestion
            errors['Warning'] = f'File has no data rows.'
```
Contextual information like filename is supplied at a higher level: it doesn't need to be repeated here. [For example](https://github.com/hubmapconsortium/ingest-validation-tools/blob/ef78c325b0e07b895f336a91c53d4089a7e55c87/dataset-examples/bad-no-such-type/README.md):
```
Metadata TSV Errors:
  dataset-examples/bad-no-such-type/submission/fake-metadata.tsv (as fake):
    Internal:
      No such file or directory: fake.yaml
    External:
      Warning: File has no data rows.
```

If you're concerned about the human-readable-flat list, that was handled in [PR #389](https://github.com/hubmapconsortium/ingest-validation-tools/pull/389#issuecomment-715609783).

General theme is that I'd prefer not to duplicate information across the response data structure.

mccalluc (changes requested):
Some comments in line. Tests fail right now:

```
Traceback (most recent call last):
  File "src/validate_submission.py", line 169, in <module>
    exit_status = main()
  File "src/validate_submission.py", line 161, in main
    errors = submission.get_errors()
  File "/home/travis/build/hubmapconsortium/ingest-validation-tools/src/ingest_validation_tools/submission.py", line 56, in get_errors
    plugin_errors = self._get_plugin_errors()
  File "/home/travis/build/hubmapconsortium/ingest-validation-tools/src/ingest_validation_tools/submission.py", line 78, in _get_plugin_errors
    for k, v in run_plugin_validators_iter(metadata_path, plugin_path):
  File "/home/travis/build/hubmapconsortium/ingest-validation-tools/src/ingest_validation_tools/plugin_validator.py", line 85, in run_plugin_validators_iter
    for k, v in validation_error_iter(data_path, assay_type, plugin_dir):
  File "/home/travis/build/hubmapconsortium/ingest-validation-tools/src/ingest_validation_tools/plugin_validator.py", line 108, in validation_error_iter
    raise ValidatorError(f'{base_dir} should be the base directory of a dataset')
ingest_validation_tools.plugin_validator.ValidatorError: /home/travis/build/hubmapconsortium/ingest-validation-tools/dataset-examples/good-scatacseq/submission/scatacseq-metadata.tsv/dataset-1 should be the base directory of a dataset
```

I'll assume that you'll fix this and subsequent errors, and ask me to re-review? Or let me know if I should help with test failures -- just don't want to step on toes.

requirements.txt
```diff
@@ -1,5 +1,5 @@
-jsonschema==3.1.0
-pyyaml==5.3.1
+jsonschema==3.2.0
+pyyaml>=5.3.1
```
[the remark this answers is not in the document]
jswelling: I feel similarly.  I tried to find a way to have my requirements.txt reference yours but I can't find a reliable way to do so.

src/ingest_validation_tools/plugin_validator.py
```diff
@@ -61,8 +62,8 @@ def run_plugin_validators_iter(metadata_path: PathOrStr,
     if metadata_path.is_file():
         try:
             with open(metadata_path, encoding='latin-1') as f:
-                rows = list(row for row in DictReader(f, dialect='excel-tab'))
-        except:
+                rows = list(DictReader(f, dialect='excel-tab'))
+        except Exception:
```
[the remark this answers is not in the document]
jswelling: Fair, but the 'open' is also in the clause.  I will use (csv.Error, IOError)

src/ingest_validation_tools/plugin_validator.py
jswelling: Yeah, we have a legitimate disagreement on this point.  My counterargument is that the additional test at this level makes the code easier to read and locally safe, which is always better than relying on a global condition which is not locally verified for code safety.  And I don't see how this 'raise' makes things *harder* to debug!  I agree that we should potentially revisit this when I've dealt with the test fixtures.

src/ingest_validation_tools/submission.py
jswelling: Fair enough, but the f'' string generates a PEP8 error from flake8 if there is no {} expression within the quoted string.  If {Path(path).name} is removed, the leading 'f' should be also.

jswelling:
OK @mccalluc I give up- how do I fix the Travis errors associated with the auto-generated docs?

mccalluc:
@jswelling : Fixed the tests locally, waiting for travis. When the fixture tests fail, there is a message telling you how to update the fixture. In this case:
```
Update needed: src/generate_dataset_docs.py codex docs/codex
Or: for D in `ls -d docs/*/  | grep -v sample`; do src/generate_dataset_docs.py `basename $D` $D; done
```
so you could run 
```
src/generate_dataset_docs.py codex docs/codex
```
or
```
for D in `ls -d docs/*/  | grep -v sample`; do src/generate_dataset_docs.py `basename $D` $D; done
```
... but it's also good to understand why there are error messages... In this case, I fixed it by adding `sort`.

>  Arguably plugin_validator.py should actually be in the ingest_validation_tests repo rather than here, because the Validator class is only used in that repo.

Thanks for pointing that out. Since nothing is calling this code right now, we don't really know whether it's working: I'd like to clarify that before merging. I may do a dummy-plugin test here, rather than worrying about the submodule... Or I may get back to that other PR...

> Perhaps the ValidationErrors should be explicitly caught at some level and converted to text error messages

I think this is the missing part that would help the re-raised exceptions make more sense to me. Let's say a file can't be read: right now, we catch that, and rethrow, and it's not bad, but we're not adding any value either. On the other hand, if at the top level we catch the errors we've raised, and update the response object instead of erroring out, then it is doing something useful. Sorry not to get where you were going earlier.

**Next**: Leaving this open a bit longer under I can exercise this code... May make a branch off of this, or merge the other PR in. Lunch first.

jswelling:
@mccalluc the current Travis errors are coming because PLUGIN_DIR_REL_PATH is invalid, and the correct value depends on where in the repo the submodule ingest-validation-tests gets mounted.  If you merge the PR that will create that submodule I'll update the PR with the appropriate value.  Alternately, if you'd prefer I can create an empty dir called dummy_plugins and point PLUGIN_DIR_REL_PATH at that.

mccalluc:
@jswelling : Thanks! I'll merge the other PR now, add a commit to set the path here, just to save back-and-forthing (hope that's ok?) and see how it looks.

mccalluc (approved):
@jswelling : Tests pass locally. Assuming they pass on travis-ci.com, I'd suggest merging this now, and then consider the next steps. I need to tackle #399 and #398. You might see other things that need to change.

mccalluc:
@jswelling ... and Travis failed ... let me see what's wrong...

jswelling:
I don't understand what's happening in the Travis case.  The code has an explicit clause to avoid importing the base class, so somehow a derived class of Validator must be getting created for which get_errors() is not overridden.  I'm sorry I don't understand the Travis environment well enough to help.  A work-around would be to point PLUGIN_DIR_REL_PATH to a directory containing no plug-ins; that should stop the plugin code from actually doing anything.

jswelling:
@mccalluc I think I found the problem.  We created the PR to pull in the ingest-validation-tests submodule when that directory was brand new, before the 'raise UnimplementedError' was added to ingest-validation-tools.  That first version has a WIP version of a test which does not implement get_errors(), so when we added the 'raise UnimplementedError' to ingest-validation-tools that version of ingest-validation-tests became poisonous.  I've added a commit to bump to the current master HEAD.

jswelling:
Aaaand now Travis is failing because the actual Validators have requirements.txt elements which are not needed elsewhere in inget-validation-tools.  I am adding them, but again, the solution may be to aim it at an empty plugin directory.  If we can get this to pass Travis I'll add a PR to make the plugin directory settable from the command line.

PR #393, opened 2020-10-23
Mccalluc/fix build

mccalluc:


mccalluc:
Travis.org is stalled, but both travis.com builds worked. merging.

Issue #399, opened 2020-10-27
allow unreferenced files.


mccalluc:

mccalluc:
@jswelling : If I remember correctly (I should have given this a better description!) the problem you're having is `No References` errors? To silence that, you could use `--submission_ignore_globs` to ignore everything. Closing now, but if I've misunderstood, please re-open with a reproducer.

PR #401, opened 2020-10-30
validate_submission.py gets command line arg to specify plugin dir

jswelling:
I think the only unexpected bit of code here is that I am treating the command line argument --plugin_dir_abs_path as an explicit request to run the plugins, so the clause skipping the plugins if there are tsv or directory errors has been removed.

### DIFF
--- a/src/ingest_validation_tools/submission.py
+++ b/src/ingest_validation_tools/submission.py
@@ -13,9 +13,6 @@ from ingest_validation_tools.validation_utils import (
 
 from ingest_validation_tools.plugin_validator import run_plugin_validators_iter
 
-# Relative path to the directory containing validation plugins
-PLUGIN_DIR_REL_PATH = "ingest-validation-tests/src/ingest_validation_tests"
-
 
 def _get_directory_type_from_path(path):
     return re.match(r'(.*)-metadata\.tsv$', Path(path).name)[1]
@@ -30,11 +27,13 @@ def _get_tsv_rows(path):
 class Submission:
     def __init__(self, directory_path=None, override_tsv_paths={},
                  optional_fields=[], add_notes=True,
-                 dataset_ignore_globs=[], submission_ignore_globs=[]):
+                 dataset_ignore_globs=[], submission_ignore_globs=[],
+                 plugin_dir_abs_path=None):
         self.directory_path = directory_path
         self.optional_fields = optional_fields
         self.dataset_ignore_globs = dataset_ignore_globs
         self.submission_ignore_globs = submission_ignore_globs
+        self.plugin_dir_abs_path = plugin_dir_abs_path
         unsorted_effective_tsv_paths = (
             override_tsv_paths if override_tsv_paths
             else {
@@ -61,7 +60,7 @@ class Submission:
         if reference_errors:
             errors['Reference Errors'] = reference_errors
 
-        if not tsv_errors and not reference_errors:
+        if self.plugin_dir_abs_path:
             # TODO: Add an option to just check plugin errors?
             plugin_errors = self._get_plugin_errors()
             if plugin_errors:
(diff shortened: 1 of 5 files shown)


src/ingest_validation_tools/submission.py
```diff
@@ -61,7 +63,7 @@ def get_errors(self):
         if reference_errors:
             errors['Reference Errors'] = reference_errors
 
-        if not tsv_errors and not reference_errors:
+        if self.plugin_dir_abs_path:
```
mccalluc, line 66:
```suggestion
        if self.plugin_dir_abs_path and not tsv_errors and not reference_errors:
```
I won't block on this, but I would prefer to keep the original checks, and I'd like to make #399 a priority, which I think would address your root problem. The problem I see is that there are a range of ways that submissions can be malformed, and these aren't "unexpected"... catching them is what this script does. The plugin checks just are more sensitive, and running them when there are more basic problems is going to cause problems.

src/ingest_validation_tools/submission.py
```diff
@@ -79,12 +81,19 @@ def get_errors(self):
...
 
     def _get_plugin_errors(self):
-        plugin_path = Path(__file__).parent / PLUGIN_DIR_REL_PATH
-        errors = defaultdict(list)
-        for metadata_path in self.effective_tsv_paths.values():
-            for k, v in run_plugin_validators_iter(metadata_path, plugin_path):
-                errors[k].append(v)
-        return {k: v for k, v in errors.items()}  # get rid of defaultdict
+        plugin_path = self.plugin_dir_abs_path
+        if plugin_path:
```
mccalluc, line 85: This is redundant with the self.plugin_dir_abs_path outside. It adds unnecessary complexity to check it inside and outside. I don't have a strong preference which one stays.

src/ingest_validation_tools/submission.py
```diff
@@ -79,12 +81,19 @@ def get_errors(self):
...
+            errors = defaultdict(list)
+            for metadata_path in self.effective_tsv_paths.values():
+                try:
+                    for k, v in run_plugin_validators_iter(metadata_path,
+                                                           plugin_path):
+                        errors[k].append(v)
+                except PluginValidatorError as e:
+                    errors['Unexpected Plugin Error'] = str(e)
```
mccalluc, line 93: This can be hit multiple times, but you'll only keep the last error. Maybe build a list of all errors?

src/validate_submission.py
```diff
@@ -157,6 +161,10 @@ def main():
         submission_args['submission_ignore_globs'] = \
             args.submission_ignore_globs
```
mccalluc, line 163:
```suggestion
```

mccalluc (commented):
I won't block if this is really the way you want it to work, but I think some changes would be good. Merge when you're happy with it, or ask for a re-review, if you think it would be helpful.

src/ingest_validation_tools/submission.py
mccalluc: (But, related to the points above, I'd prefer that the plugins not be run when their are more basic issues with the submission... so I wouldn't focus work here.)

src/ingest_validation_tools/submission.py
jswelling: Yeah, I added the second check explicitly so the impact of the conditional would be clear in the local code.  The alternative would be an 'assert self.plugin_dir_abs_path', but that would have required a try clause and thus more lines of code.

src/ingest_validation_tools/submission.py
jswelling: This error only occurs on a malfunction, rather than in response to errors in the actual submission.  Thus I thought that a single exception condition would suffice.

src/ingest_validation_tools/submission.py
jswelling: The issue is that I need to be able to *force* the plugin errors to be checked even in the presence of tsv and/or reference errors.  Would you rather have a different mechanism to force it from the command line?

mccalluc:
@jswelling : I'm going to branch from this, make it so it can run even if there are other structural errors, and get it merged, hopefully this afternoon.

mccalluc:
Extended and completed (hopefully) in #408.

PR #403, opened 2020-11-04
Update codex dir schema

mccalluc:


mccalluc:
org is stalled. merging.

PR #408, opened 2020-11-09
Mccalluc/finish plugin tests from joel

mccalluc:


mccalluc:
stalled on travis.org

PR #418, opened 2020-11-16
Add mypy

mccalluc:
There are other type checking tools, and there are a lot of options just in `mypy`, so I'm not sure if this is the best way to get started, but I think where we have annotations, they should be checked. 

I changed the imports to silence one error:
```
src/ingest_validation_tools/plugin_validator.py:119: error: Module has no attribute "util"
src/ingest_validation_tools/plugin_validator.py:120: error: Module has no attribute "util"
```
... but without the `type: ignore` it would now fail with:
```
src/ingest_validation_tools/plugin_validator.py:122: error: Item "_Loader" of "Optional[_Loader]" has no attribute "exec_module"
src/ingest_validation_tools/plugin_validator.py:122: error: Item "None" of "Optional[_Loader]" has no attribute "exec_module"
```

If I were planning to use type checking more, I would need to understand what this really means... but I'm not planning to invest time there, so an ignore seems ok?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.0.6 - In progress
 - Add thumbnail to directory schema.
 - Add machinery to include regex examples in docs.
+- Run mypy on the one file that has type annotations.
 
 ## v0.0.5 - 2020-11-09
 - Change "mixif" to "mxif".
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,3 @@
+[mypy]
+files = src/ingest_validation_tools/plugin_validator.py
+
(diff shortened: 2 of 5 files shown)


jswelling:
This looks good to me.  Am I correct that mypy is only needed in the context of testing in the ingest-validation-tools repo, so there is no need for me to add it to the requirements anywhere else?

mccalluc:
> This looks good to me. Am I correct that mypy is only needed in the context of testing in the ingest-validation-tools repo, so there is no need for me to add it to the requirements anywhere else?

Yep: It's not a run-time dependency. Behavior doesn't change with this PR.

jswelling (commented):
Fine by me.  My experience with mypy is very limited but this seems like a good place to start including it in the process.

PR #422, opened 2020-11-17
Mccalluc/first line

### LINKED ISSUE #407: Read type from first line of TSV, instead of filename

mccalluc:
Fix #407: Assay type is now read from the first line of TSVs, rather than being read from the file name, or a command line argument. Submission structures which were valid before are still valid.

@cebriggs7135: Command line interface has changed. I _strongly_ suggest that rather that trying to keep google docs in sync with this, you let me know what the needs are, and we get them addressed in this repo.

@jswelling : Would it be a plus or a minus to allow multiple TSVs of the same type to be part of a single submission? Right now if a type is repeated, only one would be seen... but it should either handle it correctly, or be an error. I'm a little leery of flexibility for its own sake... but if you have a usecase where you think splitting the TSV up would either make the your life easier, or the lives of the TMCs, I can accommodate that.


jswelling:
I see one potential problem: I believe I've seen files matching *metadata.tsv that are *not* our assay metadata in some submissions, and that they come standard with at least one assay type.  I think that if a file matching *metadata.tsv exists but does not have an "assay_type" column it should be rejected without causing an error, somewhere around submission.py:37 . Other than that it looks great to me.  I'll need to make a change to the way it's called by ingest-pipeline to match the new command line once this is merged.

jswelling:
Oh, wrt multiple files for the same assay type, I don't think that would actually break the application but I understand why you'd want to constrain the user's behavior to avoid it.  Sure, add a check that errors out when that occurs.

jswelling (changes requested):
See comments above- I think it needs to be tolerant of files matching *metadata.tsv that are unrelated to our assay metadata.

mccalluc:
@jswelling 

> I believe I've seen files matching *metadata.tsv that are not our assay metadata in some submissions, and that they come standard with at least one assay type

Can you clarify this comment? Are these files you've seen in the _submission_ directory or a _dataset_ directory? ie,
```
metadata.tsv       # Normal.
contributors.tsv   # Also typically in the submission directory, now.
weird-metadata.tsv # Is this is what you've seen ...
dataset-1/
  perhaps.fastq.gz
  or-maybe.ome.tiff
  slightly-less-weird-metadata.tsv # ... or do you mean inside a dataset?
```

If it's the latter, it's not a problem with this PR: it's only scanning the top level of the submission. If it _is_ at the top level, can you point me at an example? I'd like to understand where it's coming from.

jswelling (approved):
Your point is well taken.  I've seen experiment-metadata.tsv (or some such) in the data directories, but I can find no example of it in the top-level directory of any Release 1 dataset.  I withdraw my change request.

cebriggs7135:
@jswelling I don't know if this answers your question but the assay_metadata.tsvs are assay-specific so all the datasets listed in a given metadata.tsv are all the same assay_type.

cebriggs7135 (approved):

PR #471, opened 2020-12-10
Mccalluc/transclusion

mccalluc:
We're seeing that many of the schemas reuse swaths of columns that had previously been defined. If they should be identical across assays, it would be nice just to define it once, and then reuse. This PR demonstrates how that would work with just a couple columns. You'll also notice that little discrepancies have crept in... A period plus or minus, or a suggested default value. I would suggest that when assay-specific information is in a reused column, it be moved to the assay documentation.

- Does this make sense?
- Is this a good thing?
- Are the changes to the field descriptions acceptable?
- And can I carry this work forward?


cebriggs7135:
@mccalluc I've looked at the code changes for this PR but I don't see the demonstration of the how it would work. Can you explain?

mccalluc:
@cebriggs7135 : look at [`af.yaml` for example](https://github.com/hubmapconsortium/ingest-validation-tools/pull/471/files#diff-41a6f9b7745dd8db84c77ca894398d5884a4901b97a439432d49e9610a15651b): You can see that 4 field definitions have been removed, and replaced with a single line that includes an outside file.

cebriggs7135:
You mean like this:
 load_yaml(Path(__file__).parent / 'table-schemas' / f'{other_type}.yaml')

Yes, it is a good idea to put reusable fields in the documentation. Is that something I should help you with? Sounds labor intensive.

Screen Shot 2020-12-10 at 3 23 50 PM

mccalluc:
I don't see the relevance of the section you've quoted.

cebriggs7135:
I looked for af...yaml in the changed code and thought this is an example of what you referred to above:
"look at af.yaml for example: You can see that 4 field definitions have been removed, and replaced with a single line that includes an outside file."
That link you included didn't work so I tried to look for af...yaml in the code

cebriggs7135 (approved):

PR #472, opened 2020-12-10
test IEC examples

mccalluc:
@jswelling : No changes to the core code were required (yay!) but I have added a positive and negative test fixture, which demonstrate that these params do what you need.
```
--dataset_ignore_globs 'metadata.tsv' \
--submission_ignore_globs '*' \
--output as_text_list
```
Whether you want to use the CLI or the python interface is up to you: command line options correspond to kwargs.

Note that the flat-list-of-human-readable-errors is demonstrated here. It's been in place for a while, but I'm not sure if you all have looked at it. The code to make it human-readable is [just a list of regexps](https://github.com/hubmapconsortium/ingest-validation-tools/blob/master/src/ingest_validation_tools/message_munger.py)... If the wording isn't what you want, PRs are welcome.


jswelling (approved):
This looks good, and I think I understand the structure I need to maintain to apply this to the current real datasets.

PR #473, opened 2020-12-13
Update to describe/distinguish submission/ingest

mccalluc:
IEC transforms the data structures provided by the TMCs. This PR proposes that that structural change be part of the outward-facing documentation, and that we distinguish between a **submission structure** and an **ingest structure**. 
- I think that it is easier for the TMCs to think about a single metadata-TSV, rather than the one-line TSVs per directory...
- ... but this does have a cost, because of the mismatch between what the TMCs provide, and what IEC processes (and what will be on Globus after ingest is complete.)
- I don't think IEC will be able to change the ingest machinery to match the submission structure any time soon.

Alternatives to this PR include:
- No change to docs, but distinguish between "submission" and "ingest" in communications between IEC and Harvard: one-line TSVs for ingest are an implementation detail.
- Split this into two diagrams, instead of trying to cram into one.
- Or, keep original diagram, but add explanation to text.

Note that, as demonstrated in #472, either structure can be validated by the code. That is not in question.

I think buy-in from Joel and Nils would be sufficient, but loop in Bill and Chris if you think that's useful.


ngehlenborg (approved):

jswelling:
I'm concerned about changing the messaging we give the TMCs, especially since this process will ultimately be automated and transparent to the TMCs.  I agree that the disparity between the structure when submitted and after ingestion will differ, but the differences are small- one-line metadata.tsv files will be inserted in the data subdirectories.  In what way do we want the TMCs to change their behavior?  If we don't want them to change their behavior we shouldn't be changing the instructions.

mccalluc:
@shirey : I'm happy to drop this. I think sometimes there has been confusion about what the "submission" structure is, and this was an attempt to resolve it. If we're on the same page that the multiline-TSV format is the "real" format, and communication will be consistent, great.

The singleline-TSV format will not be purely internal, because if folks look at globus after submission, that's what they'll see, but perhaps we just ignore that?

shirey:
@mccalluc I'd leave the ingest side of things off for now.  I don't think it describes what is going to happen in enough detail- e.g. it will be a completely separate Globus endpoint where the data ends up after fully ingested.  We'll document this stuff separately for users, though, much of it I don't think will even need to be documented for end users, but more for the curation and software folks.

mccalluc:
Sounds good. Closing.

PR #475, opened 2020-12-14
Use includes

mccalluc:
Several sets of fields have been copied and pasted over time, which slight changes in wording, or typo corrections between them. By defining this in one place, we can make sure the corrections are applied consistently. 

Only the definitions are changing here: The TSVs are not touched.


cebriggs7135:
That is correct. Only the definitions are changed to make the text, punctuation, etc identical across docs.

cebriggs7135 (approved):

PR #476, opened 2020-12-14
New organs will be coming in. Loosen regex.

mccalluc:
List could be expanded to include https://github.com/hubmapconsortium/search-api/pull/224 ... but I don't think it's worth it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Update MALDI terms.
 - Demonstrate that validation of one-line-tsv-in-directory will work.
 - Add an include mechanism to reduce duplication in the configs.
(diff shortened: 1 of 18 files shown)


cebriggs7135 (changes requested):
I believe that all new tissues & organs, including Bill Shirey's list, should always be added to the search API, if I understand you correctly.

mccalluc:
@cebriggs7135 : Right, search API is expanding to handle those codes... but here in validation, the list is getting _very_ long... would you prefer that the validation here be extended to include every possible code, rather than becoming more generic?

mccalluc:
Simplifying is fine:
> OK. I see now. I thought you were talking about adding these specific terms to the menu on the left:

PR #478, opened 2020-12-14
Mccalluc/celldive

mccalluc:
One point to clarify, should there be a constraint on analyte_class? Note that this builds on the `mccalluc/more-including` branch. If you haven't already, look at that PR, too. Thanks.


cebriggs7135:
CellDIVE is antibody-based like CODEX so the analyte_class will always be proteins.

cebriggs7135 (changes requested):
CellDIVE is antibody-based like CODEX so the analyte_class will always be proteins. You can constrain analyte_type to protein(s)

PR #484, opened 2020-12-18
Ropelews/newyaml

ropelews:


ropelews:
Celldive and updated codex

mccalluc:
@jswelling : The schema is changing in a non-backward compatible way. The easiest way to see this might be to look at [this diff](https://github.com/hubmapconsortium/ingest-validation-tools/pull/484/files#diff-b8b6954f0a7f265f66e76ead44c3953116031531f6126bbd79502f5e5e24845f) ... I haven't changed the backing files for that example, and you can see how the error changed. Is this ok?

@jswelling + @ropelews : both `drv_[^/]+/channelNames\.txt` and `src_[^/]+/channelnames\.txt` are now in the schema... should both be included? and is the difference in capitalization correct?

@ropelews : Summary of the regex changes I needed to make ... mostly just syntax fixes.
```diff
-  pattern: 'HuBMAP_rounds/round_info_S[^/]\.dat'
+  pattern: 'HuBMAP_rounds/round_info_S[^/]*\.dat'
```
```diff
- .../region*/...
+ .../region[^/]*/...
```
```diff
-  pattern: 'drv_*/...'
+  pattern: 'drv_[^/]+/...'
```
```diff
-  pattern: 'src_[^/]+\/...'
+  pattern: 'src_[^/]+/...'
```
```diff
-  pattern: '*\.pdf'
+  pattern: '[^/]+\.pdf'
```
Regarding this last one, can we make this pattern tighter?

mccalluc:
(Seems to be a difference in the line order on some error messages causing test failures ... perhaps difference in python versions between local and travis? Will investigate.)

jswelling:
@mccalluc The order in which entries from a dictionary are returned by dct.items() is not deterministic in Python 2.7; unless that has changed in 3. it may be an inherent problem with the architecture of the tests.  In the past I've been forced to explicitly sort the items to get a consistent order.

jswelling:
I accept the fact that the changes may not be backward-compatible.  Since the directory structure tests were not performed by ingest-pipeline for release 1, and since the datasets supplied by the TMCs for CODEX are known to be different in this release, I don't think the lack of backward compatibility is a problem here.

jswelling (approved):
This looks reasonable to me.  I will be doing some tests against the new schemata in the near future, so getting past this PR will not be our last chance to check the consistency of these changes.

mccalluc:
The order instability comes from os.walk... so it could have been either a change in python behavior between versions, or (more likely) a difference in the FS. Sorting fixes it. FWIW: https://docs.python.org/3/whatsnew/3.7.html
> the insertion-order preservation nature of dict objects has been declared to be an official part of the Python language spec.

PR #492, opened 2021-01-13
Working to reconcile CODEX directory schema with Florida sample

jswelling:


jswelling:
@mccalluc these changes to codex.yaml were made to support submitting our sample Florida CODEX dataset with a single-line metadata.tsv in the same directory as the src_ and drv_ directories, because that's what will be needed for these interim submissions.  This is inconsistent with your test examples though.  I don't know how you'll want to resolve the differences during this period when I will be decomposing the metadata.tsv files to single lines.

mccalluc:
@jswelling :
- Adding Chris as reviewer: I don't see myself as having ownership of the content of the specs.
- I can see that you've added new requirements to the directory schema: that is fine with me, and I can update the test fixtures to match, but I don't see that it has any relationship to the 1-line tsv process, and if/when the 1-line tsv is no longer needed, and I don't think any subsequent tweaks will be required here. Am I missing something?

cebriggs7135 (approved):

jswelling:
@mccalluc thank you for cleaning up the Travis tests.  If a TMC tries to run this validation in the context of their original directory structure, that directory structure will not match this schema, correct?  Is that something we can resolve?

mccalluc:
@jswelling : See the thread in slack. Either there is some special handling that you're giving CODEX datasets which should be clarified, or it's the regular metadata splitting / TSV copying and nothing special needs to be done, apart from the ignore_globs at run time.

That concern is the only thing preventing me from approving.

mccalluc (changes requested):
The content of the PR is fine, but if this structure is different from what the TMC provides, that needs to be clarified.

ropelews (approved):
I believe these changes are consistent with our discussions.

mccalluc (approved):
We agreed that we should avoid schema changes in the future, but if a change PR does come forward, I will request @pdblood to confirm that the exception to the general policy is acceptable.

PR #503, opened 2021-01-18
Let an example be supplied to supplement regex

mccalluc:
I think there's a proposal that example filenames may be easier for the TMCs to understand that the regexes: This is one way that it could be done. The part to focus on is the `summary.pdf` below:

> |  | required? | description |
> | --- | --- | --- |
> | `[^/]+NAV[^/]+\.tif` |  | Navigational Image showing Region of Interest (Keyance Microscope only) |
> | `.+\.pdf` | ✓ | **[QA/QC]** PDF export of Powerpoint slide deck containing the Image Analysis Report |
> | `summary.pdf` | ✓ | **[QA/QC]** PDF export of Powerpoint slide deck containing the Image Analysis Report `.+\.pdf` |

I'm not sure that examples are going to be provided for everything immediately, so for now the first column might be either a pattern or an example... if it's an example, the actual pattern is relegated to the end of the description. I do check that the provided examples match the regexes.

- Perhaps less confusing to get examples in for everything? ... Though that is a significant bit of work.
- Perhaps better to be consistent and have a separate columns for example and pattern?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add encoding as CLI param.
 - Add `--offline` option, and use it internally.
 - Fix the CLI parameter parsing: Either `--local_directory` or `--tsv_paths` must be provided.
+- Allow examples of path rexes to be provided.
 
 ## v0.0.7 - 2021-01-13
 - Improved error messages in Excel.
--- a/docs/af/README.md
+++ b/docs/af/README.md
@@ -44,7 +44,7 @@ Related files:
 
 ## Directory structure
 
-| pattern (regular expression) | required? | description |
+|  | required? | description |
 | --- | --- | --- |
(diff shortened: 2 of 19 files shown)


jswelling (approved):
If this is just a change to documentation, sure.  I do think it makes things clearer for non-technical people.

cebriggs7135 (approved):

ropelews (commented):
Adding an example would be advantageous

> Perhaps better to be consistent and have a separate columns for example and pattern?
This approach would be my display preference.

mccalluc:
With the latest, if there are no examples, it will still be 3 columns, but if there is an example, it will be 4 columns. Thanks, Alex, for the nudge. To save space, I've shorted the header to just "pattern", but that could change, if needed.

PR #504, opened 2021-01-19
Mccalluc/context for decode errors

mccalluc:
Rather than requiring the TMC or the curator to count characters in the file, we can show a little bit of context in the error message. Hope this helps, @cebriggs7135?

@jswelling : My understanding is that machinery downstream hasn't really been designed with anything besides ASCII in mind, so we can make a feature request, but for now, you can't really support non ASCII TSVs, right?

(When there is time to add support, I don't have strong preferences about the encoding to use, except that it shouldn't be latin-1.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add encoding as CLI param.
 - Add `--offline` option, and use it internally.
 - Fix the CLI parameter parsing: Either `--local_directory` or `--tsv_paths` must be provided.
+- More helpful message if decoding error.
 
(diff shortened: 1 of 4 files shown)


jswelling (approved):
I *think* this makes things clearer.  At the moment, ingest-pipeline supports only ascii encoding because the downstream Neo4J interface does not seem to be able to handle more general encodings.  I agree that we will ultimately want to support a wider character set, at least for people's names.

cebriggs7135 (approved):

PR #509, opened 2021-01-21
Update antibodies.yaml

cebriggs7135:
Modified channel_id description with an example of the required format for CODEX.

### DIFF
--- a/src/ingest_validation_tools/table-schemas/antibodies.yaml
+++ b/src/ingest_validation_tools/table-schemas/antibodies.yaml
@@ -1,7 +1,7 @@
 fields:
   -
     name: channel_id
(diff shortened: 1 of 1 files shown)


mccalluc:
The channel_id should match the data inside the file, correct? If in a CODEX submission they matched, but did not match this pattern, would that be a problem?

cebriggs7135:
@mccalluc  I'll have to look at a codex dataset & the corresponding antibodies tsv to answer you. I'll also confirm with Marda so I know how to check these datasets.

mccalluc:
@cebriggs7135 -- This is replaced by #513... Beyond the structure of the ID, we need to understand how it works with data.

PR #513, opened 2021-01-25
Free text slot

mccalluc:
@cebriggs7135 : Look at [codex](https://github.com/hubmapconsortium/ingest-validation-tools/blob/40c99b8742702306cdd69e71d6402936f0e1cec1/docs/codex/README.md) ... Is that a good place to insert a note, or would you prefer it somewhere else?

@cebriggs7135 + @ilan-gold : What precisely can say about the antibody IDs in this case? What does a future developer need to know about how the ID in the antibodies TSV relates to the contents of the files?


ilan-gold:
I think using https://www.antibodyregistry.org/ was mentioned but I am not 100% sure.  Still feel like I have blinders on with all this a bit...

mccalluc:
@ilan-gold -- see slack. I'm hoping you can take the lead in asking the questions that will clarify things. (I recognize that the task is really ill-defined!)

mccalluc:
(Will learn more Friday.)

cebriggs7135:
@mccalluc @ilan-gold We know from Marda that the "cycle#_CH#" IDs in the channel_id column are generated exactly as you see them in that specific format by the Akoya instrument used for CODEX. I think I should find out what a channel_id would be for Lightsheet or MixIF or any other antibody-based assay. Then we can create more informed instructions for users in the README for each of those assays.

mccalluc:
Ilan says there will be discussion about this tomorrow. Hope to set direction after that. 

>  I think I should find out what a channel_id would be for Lightsheet or MixIF or any other antibody-based assay. Then we can create more informed instructions for users in the README for each of those assays.

That will be useful, but in this thread let's keep a focus on filling in the gaps for just this assay. Perhaps file issues for those others, if that helps? but may not be needed.

cebriggs7135:
@chuck I put assay specific channel_id into the description in the antibodies TSV bit I didn't see it show as a PR so not sure what I did wrong.

mccalluc:
@ilan-gold : Per our discussion yesterday, you'll see this through to completion? Thanks!

ilan-gold:
@mccalluc The channel ID's are basically a free text for unique identifiers.  Cytokit uses them one way (to link up to file/folder names) but I don't think there is really a catch-all for saying what they are.  For example, with assays that have fewer processing requirements, they probably are not even necessary.  I will update this PR

ilan-gold:
@mccalluc @cebriggs7135 Does this README.md for [codex](https://github.com/hubmapconsortium/ingest-validation-tools/blob/e010de5d644737aeaae88b2721e6153db9bda001/docs/codex/README.md) now look right?

mccalluc:
Thanks, @ilan-gold : 
- I'll trim off the last section about `antibody_name`,  `rr_id`, and `uniprot_id`, since what's being said will be the same for every assay type.
- I'll change the paths to globs, to make it more readable.
- and I'll see if the md generation can avoid adding lines.

cebriggs7135:
@mccalluc Going back to your original question at the top: 
 "Look at codex ... Is that a good place to insert a note, or would you prefer it somewhere else?"

I followed the link to the CODEX specs in Github but I do not see where you suggest the free text section would go. Can you help me understand better what you are asking?

mccalluc:
@cebriggs7135 
> I followed the link to the CODEX specs in Github but I do not see where you suggest the free text section would go. Can you help me understand better what you are asking?

Click on "Files changed", then go down to the go down to the readme, and click to render the markdown

https://github.com/hubmapconsortium/ingest-validation-tools/pull/513/files?short_path=ecb8358#diff-ecb8358eda22eb745f4db1748bbfe7ee7c646e9a2fc28aaec170cbf256ee7940

Is that a good place in the document to add this section, or would you prefer to see it elsewhere?

src/ingest_validation_tools/table-schemas/level-2/codex.yaml
```diff
@@ -1,4 +1,18 @@
 doc_url: https://portal.hubmapconsortium.org/docs/assays/codex
+description_md: |
+  There are 4 main fields in the antibodies.tsv file that are relevant for identifying channels within CODEX.
+  The `channel_id` is used for connecting the `antibody_name` to a given file.
+  Every `channel_id` contains a identifier for the cycle and channel number within that cycle
+  (of the form `Cycle[0-9]_CH[0-9]`)
+  while each tif file on globus contains a single channel,
```
cebriggs7135, line 7:
```suggestion
  while each tif file on Globus contains a single channel,
```

cebriggs7135 (changes requested):

PR #520, opened 2021-01-28
sample date -> sample datetime

### LINKED ISSUE #479: Answer question about sample procedure_date
This note is in the code:
>  notes: Originally, not required, but everyone filled it in. This note was present - need to be filled out only if UNET data no available (e.g. live donor). Internal value, not public, optional only for donors without unet info  - TODO - We have no machinery for keeping some fields private... Should this not be included then?

### LINKED ISSUE #507: Add time to procedure_date field in sample_metadata
@mccalluc Procedure_date should capture time of day (same as assay_metadata) because biological samples exhibit circadian effects.

https://github.com/hubmapconsortium/ingest-validation-tools/tree/master/docs/sample

mccalluc:
Fix #507; Fix #479?

If this is approved, I can then regenerate the docs, but I have several concerns:

- The request from Chris is straightforward, but it raises the bar on what we're expecting from the TMCs. Is there buy-in from every data provider?
- There was a concern expressed in the note that date alone could make the individual identifiable. Adding time would compound this. If this is approved, then I'd take it that issue (#479) is no longer a concern?
- "date" != "datetime" ... I would prefer to see this field renamed, if this goes forward.

### DIFF
--- a/src/ingest_validation_tools/table-schemas/sample.yaml
+++ b/src/ingest_validation_tools/table-schemas/sample.yaml
@@ -41,11 +41,9 @@ fields:
   name: procedure_date 
   constraints:
     required: True
-  type: date
-  format: '%Y-%m-%d'
-  description: Date of procedure to procure organ.
-  # https://github.com/hubmapconsortium/ingest-validation-tools/issues/479
-  notes: Originally, not required, but everyone filled it in. This note was present - need to be filled out only if UNET data no available (e.g. live donor). Internal value, not public, optional only for donors without unet info  - TODO - We have no machinery for keeping some fields private... Should this not be included then?
+  type: datetime
+  format: "%Y-%m-%d %H:%M"
+  description: Date and time of procedure to procure organ.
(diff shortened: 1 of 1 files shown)


ngehlenborg:
Can we add this to the agenda for our next meeting? I don't have enough context to comment.

mccalluc:
done.

cebriggs7135:
I'll ask Gloria P and Shin Lin, & TMCs re: capturing date time -sample acquisition re: pt id. TMCs have this info & how much effort?

ngehlenborg:
@mccalluc This will need to wait until we have input from HuBMAP with regard to 

1. patient identification concerns and 
2. availability of the requested information at the TMCs (we will not introduce this if it is rarely collected or available)

cebriggs7135 (changes requested):
This should not be implemented based on comments from TMCs. Time of procedure is not informative about circadian rhythm effects.

PR #521, opened 2021-01-29
Update "How to describe directories"

mccalluc:
The current documentation on the lightsheet directory structure does not have the information Alex needs to create a precise description. We will not be accepting uploads to Globus that are unvalidated, so the directory specs need to be much more precise.


cebriggs7135 (commented):
I'll ask Alex to discuss the requirements with UFLA.

cebriggs7135 (commented):
@mccalluc Alex has put this on his "urgent" list & will meet with Seth Curlin at UFLA.

cebriggs7135 (approved):

mccalluc:
@ropelews -- If this is off-base I would still appreciate your feedback, but for now, I think it will be useful to merge this.

PR #530, opened 2021-02-04
Mccalluc/versioning proposal

### LINKED ISSUE #529: Metadata schema versioning

mccalluc:
This is one way we could support versioning of metadata:
- Add an additional `version` column to the schema.
- Only one value is allowed... for the first version, it's `1` (The current state without explicit version is `0`.)
- The version of data submitted will need to match the version at the validator.
- We push tags for the earlier version to github, so the docs can link to earlier versions. (I'll make a helper script for this, if the approach seems good.)

Other possible approaches:
- Version in filename: Problematic, because we're really not constraining the names of the files... and if we did, it's not sure where it would show up downstream: The TSV filenames are not something the pipelines are capturing right now.
- Version only in TSV header, leave the cells below empty: In some sense, this would be cleaner... but it would be a little weird downstream looking at the key, rather than the value for a key.
- Version in assay name: couldn't be applied to antibodies and contributors.
- Semantic versioning: Would require more record keeping, instead of just counting down to zero to construct the tag-links for the earlier versions.
- Keep older versions in master, perhaps in separate directory: I think this would get confusing, and it would be cluttered. git provides versioning, so we should use the tools available to us.

Focus for each reviewer:
@ngehlenborg : Conceptually, does this approach seem ok?
@cebriggs7135 : For the TMCs doing data entry and reading the docs, do you think this will make sense?
@john-conroy : Code review?

Fix #529


ngehlenborg:
> We push tags for the earlier version to github, so the docs can link to earlier versions. (I'll make a helper script for this, if the approach seems good.)

So comparing v1 vs v3 would mean that the users have to jump back and forth between two versions in GitHub? I don't think that this is particularly efficient. Is there something that speaks against the docs being incremental, i.e. docs for v3 also state what is different (added, removed, changed) to v1 and v2?

Everything else is looking good.

ngehlenborg (changes requested):
> We push tags for the earlier version to github, so the docs can link to earlier versions. (I'll make a helper script for this, if the approach seems good.)

So comparing v1 vs v3 would mean that the users have to jump back and forth between two versions in GitHub? I don't think that this is particularly efficient. Is there something that speaks against the docs being incremental, i.e. docs for v3 also state what is different (added, removed, changed) to v1 and v2?

Everything else is looking good.

mccalluc:
@ngehlenborg :
> So comparing v1 vs v3 would mean that the users have to jump back and forth between two versions in GitHub? I don't think that this is particularly efficient.

I might misunderstand you, but folks could use the diff built into github, [for example](https://github.com/hubmapconsortium/ingest-validation-tools/commit/0eda0e78eee71cfd389d92f6d9d91f399297c749?short_path=c296a5e#diff-c296a5ea5161e02b39abcfea0b5186950869adf2598bd442277bb75e1aaaf331). We could generate the link to the diff, if that's a really important user story. What I want to avoid is keeping all the versions at the tip of the master branch, and I think we're on the same page about that point?

> Is there something that speaks against the docs being incremental, i.e. docs for v3 also state what is different (added, removed, changed) to v1 and v2?

Keeping a changelog is complementary to making diffs accessible. If you mean there's a changelog associated with each assay... technically, not hard, but we already have a changelog for the repo, and I'm wary of increasing the data entry requirements for Chris and Alex. 

I'll put this on the curation agenda to get on the same page.

ngehlenborg:
@mccalluc: 
 
> > So comparing v1 vs v3 would mean that the users have to jump back and forth between two versions in GitHub? I don't think that this is particularly efficient.
> 
> I might misunderstand you, but folks could use the diff built into github, [for example](https://github.com/hubmapconsortium/ingest-validation-tools/commit/0eda0e78eee71cfd389d92f6d9d91f399297c749?short_path=c296a5e#diff-c296a5ea5161e02b39abcfea0b5186950869adf2598bd442277bb75e1aaaf331). We could generate the link to the diff, if that's a really important user story. What I want to avoid is keeping all the versions at the tip of the master branch, and I think we're on the same page about that point?

Old versions of the schema don't have to be in master but the latest version of the schema should document what has been added, changed, or deprecated and in which version. For added or changed fields a comment could be added to the description ("... _Permitted values/Fieldname changed in Version 2._" or ".... _Field added in Version 4._"). For deleted fields there would have to be a separate list somewhere:

```
Deprecated Fields
 - ... (as of Version 2)
 - ... (as of Version 2)
 - ... (as of Version 3)
```

This should  be in addition to a link to the earlier version of the schema (with diff). But I am fairly confident that at least some of the submitters who create the metadata spreadsheets are not familiar with GitHub and/or diff syntax and would have a hard time even getting started finding the differences if we don't call them out explicitly.
 
> > Is there something that speaks against the docs being incremental, i.e. docs for v3 also state what is different (added, removed, changed) to v1 and v2?
> 
> Keeping a changelog is complementary to making diffs accessible. If you mean there's a changelog associated with each assay... technically, not hard, but we already have a changelog for the repo, and I'm wary of increasing the data entry requirements for Chris and Alex.

So this is not an explicit CHANGELOG (again, not sure if everyone would know to look there for changes) but the docs contain the critical information.

> I'll put this on the curation agenda to get on the same page.

It would still be good to discuss this in person.

john-conroy (approved):

mccalluc:
That's doable: Latest version has output that looks like:

> Changes:
> - v1: Version number added.
>
> Previous versions:
> - [v0](https://github.com/hubmapconsortium/ingest-validation-tools/tree/antibodies-v0/docs/antibodies) / [diff](https://github.com/hubmapconsortium/ingest-validation-tools/compare/antibodies-v0...master)

The "Changes" section is free text: Chris or whoever makes a change will be responsible for updating it. The previous versions links are generated. (Linking to the diff of just the README is not feasible.)

The "Change" section could be more structured... but that makes the data entry more complicated, and it will probably devolve on me to keep it organized.

ngehlenborg (approved):

cebriggs7135 (approved):

cebriggs7135:
I've been reading up on HCA metadata specs like versioning:
https://data.humancellatlas.org/metadata/rationale#design-choices 
"Currently, the schema name and version is declared in the describedBy field using a URL. This approach allows a metadata consumer to fetch the schema associated with a metadata document."

mccalluc:
@cebriggs7135 -- Anything is possible, but let's not discuss it in a closed PR. Separate issue / slack / meeting agenda would be better.

PR #533, opened 2021-02-08
Update celldive.yaml

cebriggs7135:
channel_id for CellDIVE description text.


src/ingest_validation_tools/table-schemas/level-2/celldive.yaml
```diff
@@ -1,6 +1,17 @@
 # doc_url: TODO https://github.com/hubmapconsortium/ingest-validation-tools/issues/480
 fields:
 # Overrides of Level 1 fields:
+-
```
mccalluc, line 4: This should be `description_md`. I'd also prefer that this only describes the parts that are different for this assay. I'll edit this, and ask for you to check it.

mccalluc (commented):
I'll make changes.

PR #534, opened 2021-02-09
no more stand-alone TSV validation

mccalluc:
The stand-alone TSV validation is not part of any workflow now, and I think its continued availability will cause confusion: It makes it unclear what "validated" means. If someone *really* needs to validate an isolated TSV, they can drop it into an empty directory and validate that, and only look at the errors from the TSV.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Add flowchart documenting the consensus submission process.
 - cleanup-whitespace.py
 - Issue templates to operationalize new process for handling post-release changes.
+- Remove option to just validate TSVs.
 
 ## v0.0.7 - 2021-01-13
 - Improved error messages in Excel.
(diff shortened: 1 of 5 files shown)


cebriggs7135:
@mccalluc  Even sample metadata tsv validation? Sample_metadata is not part of a submission directory and I don't think you validate it other than as a stand alone. Is that correct?

jswelling (commented):
I don't think this is a good idea.  Stand-alone validation is not part of *our* workflow, but it may very well be part of *their* workflow.  I can easily see a scenario where one contributor builds the .tsv files from Excel and a different contributor moves the data files into place on the host from which they will eventually be pushed to HIVE.  The .tsv contributor will want to check that they have produced syntactically correct files before handing them off to the second contributor for inclusion in the upload.
I am not going to 'request changes' but I think these scenarios should be discussed more broadly.

mccalluc:
@jswelling - That's a really good point. Closing.

PR #549, opened 2021-02-16
Add version 1 everywhere

mccalluc:
One question: table of contents for each assay currently is grouped, with "Provenance" fields first... Should `version` be added to that, or stand on its own? [For example](https://github.com/hubmapconsortium/ingest-validation-tools/blob/813beddbc1c965cb263c194d8de1a0824a5245b9/docs/af/README.md#table-of-contents).

Draft for now: I want to add more testing so error messages when it's missing will be clear.


cebriggs7135:
@mccalluc I think it's correct to have it stand alone because it's not related to provenance.

ngehlenborg (approved):
I agree that separate from provenance is preferable.

mccalluc:
Joel has said adapting to this when older TSVs are still not ingested would be disruptive. Placing on hold until there's a policy on change management that is acceptable to all parties. On agenda for February 19.

mccalluc:
Consensus on call was that older versions still need to be validatable on HEAD. That requirement is incompatible with this approach. Closing.

PR #553, opened 2021-02-16
Mccalluc/concentration fields

### LINKED ISSUE #550: New antibodies version with concentration fields
Add a column for a concentration_value and concentration_unit. Concentration_unit should be a constrained to "ug/ml".

mccalluc:
@cebriggs7135 : Can you use the github suggest tool to fill in the descriptions on the yaml? Thanks

Fix #550


src/ingest_validation_tools/table-schemas/antibodies.yaml
```diff
@@ -32,6 +33,18 @@ fields:
       required: False
       pattern: '1/\d+'
     example: 1/200
+  -
+    name: concentration_value
+    description: TODO!
```
cebriggs7135, line 38:
```suggestion
    description: The concentration value of the antibody preparation.
```

src/ingest_validation_tools/table-schemas/antibodies.yaml
```diff
@@ -32,6 +33,18 @@ fields:
...
+    type: number
+    constraints:
+      required: False
+  -
+    name: concentration_unit
+    description: TODO!
```
cebriggs7135, line 44:
```suggestion
    description: The concentration units of the antibody preparation.
```

cebriggs7135 (changes requested):

src/ingest_validation_tools/table-schemas/antibodies.yaml
cebriggs7135: @mccalluc  concentration units should be optional, as the value is.

cebriggs7135 (changes requested):

src/ingest_validation_tools/table-schemas/antibodies.yaml
mccalluc: > @mccalluc concentration units should be optional, as the value is.

Sorry, I missed this comment. I don't believe we should do it that way: We do not have a mechanism to make one field required only if another is present. Between having a default unit in the column, when there's no value for it to apply to, or missing the unit when it is needed, I prefer the former.

PR #555, opened 2021-02-18
Mccalluc/factor out table validation

mccalluc:
Could one of you review this? Whoever notices it first...

I want to pull out and test the table validation on its own, because there is upcoming work that is limited to this interface. Right now, we have coverage end-to-end, but it depends on a lot of other details, like the loading of assay-type specific table schemas, so it's hard to test and make improvements limited to the table validation.

One implementation choice that could be questioned: Writing tests in bash instead of with pytest. I think file-based fixtures make sense here, using either approach, and I could configure pytest to load in fixtures from a directory... but assuming some familiarity with bash, this is easier to write and read without consulting the manual. Downsides: 
- Slower (need to start-up process for each fixture)
- Reporting (Using pytest would give us consistent reporting across all tests)
- Extras (Pretty easy to parallelize pytest tests)

Please push back if you don't think this is a good approach!


john-conroy (approved):
It shouldn't influence direction, but I'm more familiar with pytest than bash and while we don't expect me to touch these tests, it could impact the quality of my reviews.

PR #563, opened 2021-02-19
Update to IMC yaml- removed 2 redundant datetime fields

cebriggs7135:
The IMC schema includes 2 datetime fields that are not necessary. The execution_datetime field captures the start_datetime and the end_datetime is not relevant. I have removed these 2 fields:

start_datetime and end_datetime
Per 2/19/21 conversation with the IMC team from UFLA/Zurich:


ngehlenborg (approved):

mccalluc:
Fixed by #635

PR #564, opened 2021-02-19
Update to IMC yaml: Add antibodies_path

cebriggs7135:
IMC uses antibodies. An antibody_path field should be added.


mccalluc:
Fixed by #635

PR #565, opened 2021-02-22
declarative network checks

### LINKED ISSUE #495: Make network checks declarative
Rather than ad-hoc code, consider registering an additional check with table-validations.

mccalluc:
All checks for IDs (protocols/rr_id/orcid/...) are now part of the respective schemas, and validated with an extension to the table validator. Advantages:
- ID validation is done consistently for all cases.
- New ID validations can be done in schema, rather than writing new python code.
- Since it's now part of the table definition, auto-generated documentation can explain these constraints to users.

This is also a test case for writing validation extensions: I wanted to start by translating something we already have into this new idiom, rather than starting something entirely new -- If there are problems with the extension machinery itself, more likely to be caught this way.

The wording of some error messages has changed, but whether a particular submission passes or fails validation should not.

Fix #495


cebriggs7135 (approved):

jswelling (approved):
Looks like a good idea, but I'm going to have to try it out with real data before I really understand it.

PR #566, opened 2021-02-23
Mccalluc/units for

mccalluc:
@john-conroy : code review
@cebriggs7135 : See annotations on the individual files in PR: comments will be easier to understand in context.


dataset-examples/good-codex-akoya/submission/antibodies.tsv
```diff
@@ -1,2 +1,2 @@
 version	channel_id	antibody_name	rr_id	uniprot_accession_number	lot_number	dilution	concentration_value	concentration_unit	conjugated_cat_number	conjugated_tag
-2	channel-depends-on-assay	whatever	AB_1934296	G9N9I7	whatever	1/100	""	ug/ml		whatever
\ No newline at end of file
+2	channel-depends-on-assay	whatever	AB_1934296	G9N9I7	whatever	1/100				whatever
```
mccalluc, line 2: Demonstrating that the unit can be left off if there is no value.

src/ingest_validation_tools/table-schemas/antibodies.yaml
```diff
@@ -51,6 +51,9 @@ fields:
     constraints:
       enum:
         - ug/ml
+      required: False
```
mccalluc, line 54: We'll need to explicitly mark unit fields so they are no longer required.

table-examples/units-for/output.txt
```diff
@@ -0,0 +1,2 @@
+On row 3, column "a_unit", value "None" fails because Required when a_value is filled
+On row 4, column "a_unit", value "None" fails because Required when a_value is filled
```
mccalluc, line 2: Most of the text of this message is generated in code outside this PR... so it could change, if needed, but that's a larger scope.

cebriggs7135 (approved):

john-conroy (approved):

PR #568, opened 2021-02-23
Flag fields which have been auto-incremented by Excel

### LINKED ISSUE #540: Help users if they've autofilled protocol IDs, and the numbers have incremented.
@cebriggs7135 :  Do we know that TMCs are actually unable to interpret the error? ... or is the error message just too verbose, but still intelligible?

mccalluc:
Fix #540 

Schemas now can specify the length of sequence of incrementing integers which will be flagged as an error, because it's probably coming from Excel. This is different from the other validations because it is at the core a heuristic, so the limit should be set conservatively, since we don't have a distinction between non-fatal warnings, and fatal errors.

This could be exposed in the generated documentation, but I don't think that's necessary: This doesn't change anything about the *meaning* of the fields.

@cebriggs7135 -- I think it will be your responsibility to decide what the limit should be, where it should be applied.

(Not sure why mypy was scanning this file: I'm not interested in diving into those annotations right now and I just want to silence the error.)


cebriggs7135 (changes requested):
The cutoff should be the first instance of an increment of 1. Any increment should be flagged for review but not necessarily as an error that must be changed - in case it's correct.

mccalluc:
@cebriggs7135 -- We have no machinery for warnings which are not errors.... and I'm reluctant to add it: We would need a way to configure warnings separately from errors, and we would need a way to report them separately. These are the ways I see forward:
- Merge this, but use high limits, so that only situations which are definitely errors will be flagged.
- Merge this, and configure it to error after runs of only 2: The TMC will be responsible for reordering the rows to avoid sequences.
- Wait until "warnings" are handled separately from "errors". (This will not come for a long time.)

cebriggs7135 (approved):

jswelling (approved):
I concur that we can flag it as an error after a run of only 2, because re-ordering the lines is always an option.  Well, I can think of a really convoluted example where reordering won't work, but the option of splitting the .tsv file would still exist.

PR #569, opened 2021-02-23
Updated RNA-Seq directory structure schema YAML file

### LINKED ISSUE #452: Directory schema for scrnaseq
This directory structure schema will apply to these assay-types

- scRNAseq-10xGenomics
- scRNAseq
- sciRNAseq
- snRNAseq
- SNARE2-RNAseq

icaoberg:
Updated scRNA-Seq directory schema after testing and validating the folder structure on a sample dataset from a data provider. Used the protected dataset with UUID `638799...` to test schema.

## Comments

* There seems to be a pattern in the folder and filenames associated with this assay type. However I don't have documentation stating these patterns must be verified.
* Need to include an assay type dataset description at the top of the YAML file. Maybe @cebriggs7135 or @pdblood have a short description that I can add to the file?

If approved, this resolves #452.

### DIFF
--- a/src/ingest_validation_tools/directory-schemas/scrnaseq.yaml
+++ b/src/ingest_validation_tools/directory-schemas/scrnaseq.yaml
@@ -4,3 +4,8 @@
   pattern: '[^/]+\.fastq\.gz'
   description: 'Compressed FastQ files'
   required: True
+
+-
+  pattern: '[^/]+/[^/]+\.fastq'
+  description: 'FastQ files'
+  required: False
--- a/src/ingest_validation_tools/directory-schemas/scrnaseq.yaml
+++ b/src/ingest_validation_tools/directory-schemas/scrnaseq.yaml
@@ -1,6 +1,6 @@
 #scRNA-Seq description
 
 -
-  pattern: '[^/]+/[^/]+\.fastq\.gz'
+  pattern: '[^/]+\.fastq\.gz'
   description: 'Compressed FastQ files'
   required: True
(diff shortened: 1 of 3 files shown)


src/ingest_validation_tools/directory-schemas/scrnaseq.yaml
```diff
@@ -1,4 +1,6 @@
--
-  pattern: 'TODO'
-  description: 'Directory structure not yet specified. https://github.com/hubmapconsortium/ingest-validation-tools/issues/452'
+#scRNA-Seq description
 
+-
+  pattern: '[^/]+/[^/]+\.fastq\.gz'
```
mccalluc, line 4: @icaoberg -- Are you confident about this? Do we need to require in each dataset (as opposed to the submission) that there be one directory level above the `.fastq.gz`? If that's really the case, great, but I would be less surprised by either 
```suggestion
  pattern: '.+\.fastq\.gz'
```
ie, allow files with any directory structure, or
```suggestion
  pattern: '[^/]+\.fastq\.gz'
```
ie, `.fastq.gz` must be at top level.

mccalluc (commented):
See comment. I'll rebuild the docs and push to this branch.

mccalluc:
Actually... master seems to be broken... investigating...

icaoberg:
@mccalluc please let me know if you have any questions about this PR. The directory structure schema for `scatacseq` should be the same as this one. So any question I reply about the directory structure on the other ticket applies here as well.

mccalluc:
Replaced by #587

PR #575, opened 2021-02-24
Update sample TSV: add 4C to sample preservation temp

cebriggs7135:
UFL uses 4C for sample preservation temp.


src/ingest_validation_tools/table-schemas/sample.yaml
```diff
@@ -98,6 +98,7 @@ fields:
       - Freezer (-80 Celsius)
       - Freezer (-20 Celsius)
       - Room Temperature
+      - 4C
```
ngehlenborg, line 101:
```suggestion
      -  Refrigerator (4 Celsius)
```

Is that correct? The options should follow the same pattern if possible.

ngehlenborg (changes requested):

src/ingest_validation_tools/table-schemas/sample.yaml
cebriggs7135: Sounds good.

src/ingest_validation_tools/table-schemas/sample.yaml
```diff
@@ -98,6 +98,7 @@ fields:
       - Freezer (-80 Celsius)
       - Freezer (-20 Celsius)
       - Room Temperature
+      -  Refrigerator (4 Celsius)
```
mccalluc, line 101:
```suggestion
      - Refrigerator (4 Celsius)
      - Room Temperature
```
Swapping the order, just so they are sorted correctly.

PR #577, opened 2021-02-25
Downgrade requests

mccalluc:


jswelling (approved):
That's it.  I've tested it in this configuration and it seems to work just fine.

PR #581, opened 2021-02-25
Updated ATAC-Seq directory structure YAML file

icaoberg:
This directory structure schema applies to these assay-types

 - SNARE-seq2
 - scATACseq
 - sciATACseq
 - snATACseq


jswelling (approved):
Looks good to me!

mccalluc (commented):
@icaoberg - Thanks! I'll regenerate the docs before merging. 

Can you explain the directory structure?
```
tmc-submission/
  tmc-dataset-1/
    subdirectory-1/
      one.fastq.gz
    subdirectory-2/
      two.fastq
```
Does `subdirectory-1` vs `subdirectory-2` have any significance to the data provider? To our pipelines?

(This type was used in fixtures, so I've added the extra directories which are no required -- If this is correct, great, but I'd like to understand where they're coming from.)

Also: Any chance they should both be optional? If someone only submitted `.fastq` (not `.fastq.gz`) that would fail -- That's the correct behavior?

icaoberg:
@mccalluc the expected directory structure is meant to look like this

```
tmc-submission/
  ...*.tsv
  tmc-dataset-1/
      one.fastq.gz
      two.fastq
    tmc-dataset-2/
      one.fastq.gz
      two.fastq
```

for a submission that includes multiple datasets. When split, each dataset should look like 

```
  ...*.tsv
  tmc-dataset-1/
      one.fastq.gz
      two.fastq
```

and

```
  ...*.tsv
  tmc-dataset-2/
      one.fastq.gz
      two.fastq
```

I think the expectation is that people will submit compressed FastQ files. If this makes validation easier, that is fine with me. However @mruffalo and @jswelling should comment on this. Do the pipelines expect `.fastq.gz`, `fastq` or either?

mccalluc:
Talked with Ivan: There was a misunderstanding about the appropriate level to take as the base in the directory description. Taking Ivan's example:
```
tmc-submission/
  metadata.tsv
  tmc-dataset-1/
      one.fastq.gz
      two.fastq
```
the post-split dataset will _actually_ be 
```
  tmc-dataset-1/
      one-line-metadata.tsv
      one.fastq.gz
      two.fastq
```
In either case, the `fastq` files are at the _top level_ of the dataset, and the appropriate regex is just `[^/]+\.fastq.gz`

We agreed it would be easiest just to close this (since I added all those changes to the fixtures) and start over. @jswelling -- Since you approved this, does what I say make sense? Or am I really confused? Thanks!

PR #585, opened 2021-02-26
Mccalluc/atacseq dir

### LINKED ISSUE #451: Directory schema for scatacseq

mccalluc:
Replacement for #581, after talking with Ivan. Because both patterns are optional (a dataset could be only `.fastq` or only `fastq.gz`), the missing file warning is longer generated in a couple test fixtures.

... if some warning should be presented, the pattern could be changed to 
```
[^/]+\.fastq(\.gz)?
```
... but I think the status quo is fine.

Fix #451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fix typo in nano enum.
 - Clearer error when it can't find matching assay name.
 - Downgrade dependency for compatibility with HuBMAP commons.
+- Directory structure for scatacseq.
 
 ## v0.0.8 - 2021-02-10
 - Update CODEX directory structure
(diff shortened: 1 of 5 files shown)


icaoberg (approved):
Looks good to me.

@mccalluc I noticed you remote my comment header. That's fine with me. If you don't want me to include them in the future, just let me know.

jswelling (approved):
I concur, since both are optional.  It would be nice if we had a way to say 'one or the other must exist' though.

mccalluc:
@icaoberg : 
> I noticed you remote my comment header. That's fine with me. If you don't want me to include them in the future, just let me know.

Let me add that back in ... I have no objection, but we need to remember that it's really only going to be visible to developers ... If the information should have a broader audience, we should try a different approach.

@jswelling :
>  It would be nice if we had a way to say 'one or the other must exist' though.

changing the regex to `[^/]+\.fastq(\.gz)?`, and making the single line required, would accomplish that. I won't make that change myself (prefer not to wear the schema-writer hat!), but PRs are welcome if that seems like the right direction.

PR #592, opened 2021-03-01
Mccalluc/3d imc

mccalluc:
- @cebriggs7135 - Check that the list of fields is correct.
- @jswelling - This introduces a new assay type: Now that there is the assay service, what should I do to try to bring this in sync with that? Does there need to be a corresponding PR somewhere else?


jswelling (commented):
The new type will need to be added to https://github.com/hubmapconsortium/search-api/blob/master/src/search-schema/data/definitions/enums/assay_types.yaml .  I'm not clear on how you're supposed to create a branch and a PR to integrate with Bill's current deployment process.

cebriggs7135:
@mccalluc  You say above " factor out roi fields". That sounds like you've removed them. That's incorrect. The roi fields remain in 3D IMC.

mccalluc:
> That sounds like you've removed them.

@cebriggs7135 -- No: That's a misunderstanding. "factoring out" is a metaphor, based on the operation in algebra:
```
ax + bx = (a + b)x
```
... except in this case, we can think of `ax` as being the IMC fields, `bx` as being the 3d IMC fields. I've pulled out the parts they two assays share, and reference these shared parts from each assay.

You can see that there are no changes to imc.md in this PR. I would like you to review imc3d.md, to confirm that it has the correct fields.

mccalluc:
> The new type will need to be added to ...

Filed https://github.com/hubmapconsortium/search-api/pull/237 ... I don't think it's necessary to wait until that is merged to merge this, but defer to your judgement.

cebriggs7135 (approved):

mccalluc:
@jswelling - ping - https://github.com/hubmapconsortium/search-api/pull/237 is open - I don't think this PR needs to block on that, but I'll defer to you.

jswelling (approved):
Looks fine, but directory schema still needed.

PR #600, opened 2021-03-04
make as_text_list the default

### LINKED ISSUE #596: Make flat list the default output format
Chris concurs with Bill that this will be easier for end-users to read. Exposing regexes is fine, since we don't have examples for most fields.

mccalluc:
I remembered to ask Chris, and she concurs with Bill that the flat list is the easier format for typical end users to read. I'm not really convinced of this, but I'm also not the best judge, so this makes it the default. No tests needed to be changed, because nowhere were we just using the default: Instead, `as_md` was used everywhere. I've removed `as_flat_list` in one test, because it is now the default.

I'm not sure what Airflow uses, but it should probably be consistent?

fix #596 

cc @cebriggs7135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Downgrade dependency for compatibility with HuBMAP commons.
 - Directory structure for scatacseq.
 - Factor out the checks, make OO, and make error messages configurable.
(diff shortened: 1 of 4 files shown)


jswelling (approved):
Makes sense to me.

Issue #601, opened 2021-03-04
Uniform handling of errors from dict_reader_wrapper


mccalluc:
It's called from a number of locations in the code, and different errors are trapped in different locations: The error trapping should be handled in one spot, and just a single error should be re-thrown, and callers will be responsible for trapping just that one error.

PR #602, opened 2021-03-04
Mccalluc/dot for contributors

### LINKED ISSUE #595: Friendlier errors on TSVs with blank lines
- if you reference a directory where it expects a file, give an error instead of a stack trace…
- if a line is simply empty, error out sooner, instead of trying to parse it.

<details><summary>stack trace</summary>

```
src/validate_submission.py --local_directory /tmp/hunter
Traceback (most recent call last):
  File "src/validate_submission.py", line 155, in <module>
    exit_status = main()
  File "src/validate_submission.py", line 147, in main
    errors = submission.get_errors()
  File "/Users/chuck/github/hubmap/ingest-validation-tools/src/ingest_validation_tools/submission.py", line 99, in get_errors
    tsv_errors = self._get_tsv_errors()
  File "/Users/chuck/github/hubmap/ingest-validation-tools/src/ingest_validation_tools/submission.py", line 152, in _get_tsv_errors
    self._get_single_tsv_external_errors(assay_type, path)
  File "/Users/chuck/github/hubmap/ingest-validation-tools/src/ingest_validation_tools/submission.py", line 192, in _get_single_tsv_external_errors
    contributors_path)
  File "/Users/chuck/github/hubmap/ingest-validation-tools/src/ingest_validation_tools/submission.py", line 214, in _get_contributors_errors
    encoding=self.encoding, offline=self.offline)
  File "/Users/chuck/github/hubmap/ingest-validation-tools/src/ingest_validation_tools/validation_utils.py", line 73, in get_internal_errors
    rows = dict_reader_wrapper(path, encoding)
  File "/Users/chuck/github/hubmap/ingest-validation-tools/src/ingest_validation_tools/validation_utils.py", line 19, in dict_reader_wrapper
    with open(path, encoding=encoding) as f:
IsADirectoryError: [Errno 21] Is a directory: '/tmp/hunter'
```

</details>

<details><summary>chat</summary>

```
Me to Hunter Hakimian (Direct Message) (11:41 AM)
Ok… Think I know what’s going on, but it’s a case I hadn’t considered, and the error reporting needs to be better …
There are a lot of blank lines at the end of the TSV…
But it’s still trying to read them …
Hunter Hakimian to Me (Direct Message) (11:42 AM)
I see. I'm not sure why those are there
Me to Hunter Hakimian (Direct Message) (11:43 AM)
And so when it sees an empty value in antibodies.tsv or contributors.tsv it resolves to the directory…
And since it’s expecting a file, it gives the error you see…
So, work on my side:
Hunter Hakimian to Me (Direct Message) (11:43 AM)
I see
Me to Hunter Hakimian (Direct Message) (11:44 AM)
- if you do reference a directory where it expects a file, give an error instead of a stack trace…
- if a line is simply empty, error out sooner, instead of trying to parse it.
Hunter Hakimian to Me (Direct Message) (11:45 AM)
Ok, I'll try to find a way to clean this file way up on my end.
Me to Hunter Hakimian (Direct Message) (11:46 AM)
There’s an argument for making the parsing more robust … but there are several different systems downstream from here, so we’ve thought it will cause the fewest headaches to be strict when things come in.
```

</details>

### LINKED ISSUE #597: Supplying a directory path rather than a file for contributors_path causes a stack dump from validate_submission
Tested at c60774c45fc0477bcd1 on master.  To duplicate, take a valid metadata.tsv file and replace the contributors_path element with just '.' .  This is reported as a bug on the theory that stack traces need to produce output error messages instead.

[stack_trace.txt](https://github.com/hubmapconsortium/ingest-validation-tools/files/6072145/stack_trace.txt)

mccalluc:
Fix #597 and fix #595: Add a test case that demonstrates both kind of problems.

Implementing this, realized that everywhere the helper is called, the exception trapping is being done differently. It needs to be moved inside: Filed #601 

@cebriggs7135 : Main thing for you to notice is the kind of error produced in these situations, so you can point the TMCs on the right track, if they're confused.


jswelling (approved):
Ooh, I learned a new built-in exception!

PR #606, opened 2021-03-06
versioning

mccalluc:
Enable versioning of schema. Now, when a TSV is read, it reads not only the assay type, but also the version to determine the schema to validate against. The most significant thing about this PR is that the existing submission fixtures did not need to be touched: if it failed before, it still fails, with the same message, and if it passed before, it still passes. Worth checking out the generated MD and making sure it's what you want: [for example](https://github.com/hubmapconsortium/ingest-validation-tools/blob/217617965e3cd0bcb329ae4a08a09aacae9f4728/docs/scatacseq/README.md)
- Each version is included in a collapsable `<details>` section: only the most recent is expanded.
- The Excel and TSV templates are only available for the most recent version.

Also:
- A lot of the code changes involve moving towards more uniformity between the handling of assays and the handling of others. So much has moved around that the diffs aren't very readable. 
- The earlier take on this involved git tags, and that is no longer the case: Everything is at the head on master.
- Only antibodies and scatacseq are have multiple versions right now: I didn't want to overwhelm this PR with even more details.


jswelling:
@mccalluc I'd like to look at how validate_submission decides which version of the schema to apply to a given metadata.tsv .  Given all the diffs, can you point me to the block(s) of code that implement this?

mccalluc:
@jswelling - There are actually two points where it reads the version number: At the two points where it used to just read the `assay_type`, now it also reads the `version`, but for right now, after it reads the `version` of the first line in [submission.py](https://github.com/hubmapconsortium/ingest-validation-tools/pull/606/files#diff-dc4e635d59baa070134a4f72ffa30c52649f079bcd4f8364a2e19353deb38a78R95) it only keeps the `assay_type`, and needs to re-read the `version` in [validation_utils.py](https://github.com/hubmapconsortium/ingest-validation-tools/pull/606/files#diff-4eac397eb6d8f2e4b35ef29e5c2d37f6c26bf44218e9bdb76ff288c128ff3739R85).

I would like to refactor it...
- Only reading those fields at the low level, when the whole file is about to be read.
- Centralizing error handling (#601)
- Use consistent variable names, and distinguish between `assay_name` (the value in the TSV) and `schema_name` (the name of the yaml, without the version or the extension.)

... but adding more tweaks in this branch would bog it down, I'm afraid.

cebriggs7135 (approved):

jswelling:
So this would invalidate all existing .tsv files?  Is there a default version number?

mccalluc:
@jswelling -
>So this would invalidate all existing .tsv files? 

No: You can see that the submission text fixtures were not touched in this PR, and the `good-*` fixtures are still good.

> Is there a default version number?

yes: zero

Please look at the links I provided in response to your earlier question:
```
    version = rows[0]['version'] if 'version' in rows[0] else '0'
```

ngehlenborg (changes requested):
I am struggling with the layout of the rendered markdown pages. I don't understand the hierarchy of the content. Are "version 1" and "version 2" for the directory schema in this example? 

Does should "Table of Contents" be at the same level as "Directory Structure"?

Maybe you try to style the headings in the markdown so that the structure becomes more obvious? 

image

mccalluc:
@ngehlenborg -- would this be better?

The headings on the page would be:
```
# scatacseq
## Directory schema
## Metadata schema
### Older versions
### Current version: 2
         Summary goes here (was ToC)
#### Shared by all types
##### `field_1`
##### `field_2`
#### Unique to this type
##### `field_1`
##### `field_2`
```

- Give "Metadata schema" an H2 header, with H3s for "Older versions" and "Current version" under it.
- "Table of Contents" is replaced by a summary without its own header, just under the header for the "Current version".
- Drop level-1/level-2, and instead call it what it is: "Shared by all types", "Unique to this type"
-----

# scatacseq

Related files:
- [🔬 Background doc](https://portal.hubmapconsortium.org/docs/assays/atacseq): More details about this type.
- [📝 Excel template](https://raw.githubusercontent.com/hubmapconsortium/ingest-validation-tools/master/docs/scatacseq/scatacseq-metadata.xlsx): For metadata entry.
- [📝 TSV template](https://raw.githubusercontent.com/hubmapconsortium/ingest-validation-tools/master/docs/scatacseq/scatacseq-metadata.tsv): Alternative for metadata entry.
- [💻 Metadata schema](https://github.com/hubmapconsortium/ingest-validation-tools/edit/master/src/ingest_validation_tools/table-schemas/assays/scatacseq.yaml): To update metadata fields.
- [💻 Directory schema](https://github.com/hubmapconsortium/ingest-validation-tools/edit/master/src/ingest_validation_tools/directory-schemas/scatacseq.yaml): To update directory structure.

## Directory schema

| pattern | required? | description |
| --- | --- | --- |
| `[^/]+\.fastq\.gz` | ✓ | Compressed FastQ file |
| `extras/.*` |  | Free-form descriptive information supplied by the TMC |
| `extras/thumbnail\.(png\|jpg)` |  | Optional thumbnail image which may be shown in search interface |


## Metadata schema
### Older versions

<details ><summary>Version 0</summary>
all the fields
</details>

<details ><summary>Version 1</summary>
all the fields
</details>

### Current version: 2

<details><summary>Shared by all types</summary>

[`donor_id`](#donor_id)<br>
[`tissue_id`](#tissue_id)<br>
etc.

</details>

<details><summary>Unique to this type</summary>

[`execution_datetime`](#execution_datetime)<br>
[`protocols_io_doi`](#protocols_io_doi)<br>
etc.

</details>

#### Shared by all types

##### `donor_id`
HuBMAP Display ID of the donor of the assayed tissue. Example: `ABC123`.

| constraint | value |
| --- | --- |
| pattern (regular expression) | `[A-Z]+[0-9]+` |
| required | `True` |

etc.

#### Unique to this type

##### `something_else`
Blah blah blah. Blah blah blah. Blah blah blah. Blah blah blah.

| constraint | value |
| --- | --- |
| pattern (regular expression) | `[A-Z]+[0-9]+` |
| required | `True` |

etc.

ngehlenborg:
Yes, thank you. That is much clearer.

I would suggest to invert the order of the versions (start with most recent, end with oldest) and maybe change the titles to "Version 2 (current)" and then have a section below "Previous Versions" with subsections "Version 1" and "Version 0".

mccalluc:
Everything resolved! Again, this doesn't actually create a v1 for most types, but there are examples to show that it does work, and the work can continue in a new branch.

PR #611, opened 2021-03-09
Mccalluc/generate another report

### LINKED ISSUE #610: Collect values of free text fields... could they map to ontologies?
To support moves towards ontologies, gather values of all free text fields, along with the field and data type they come from, and deliver to Jonathan.

mccalluc:
Fix #610


mccalluc:
TODO: 
- remove numeric values and dates... remove values from report if there is not one alphabetic character.
- Make sure Sample metadata is covered in the report.
- Add assay_name to report

Share results on curation channel, not with Jonathan.

PR #612, opened 2021-03-10
Fixed two bugs in the 3D IMC directory structure YAML that were preventing local validation

icaoberg:
This PR fixes two issues that were found when UFL was attempting to validate locally.

* Folder is meant to be named `Mapping` not `mapping` (case sensitive)
* Zip files with contain the name of the organ, not the word `organ`.


src/ingest_validation_tools/directory-schemas/imc3d.yaml
```diff
@@ -17,6 +17,12 @@
   required: True
   is_qa_qc: True
 
+-
+  pattern: 'mcd/panel\.csv'
+  description: 'CSV containing a list of channels and antibodies used for acquisition'
```
mccalluc, line 22: How is this different from the antibodies TSV?

mccalluc (commented):
@icaoberg
- I'm hesitant about the `panel.csv` ... This seems to be exactly what the `antibodies.tsv` is for. 
- Not a blocker, but is there any chance we can push for lower case filenames? This might not be the place... but I hope there are some cases where we can steer the TMCs to make less arbitrary decisions.

src/ingest_validation_tools/directory-schemas/imc3d.yaml
icaoberg: @mccalluc I have no clue. this is a question for @cebriggs7135 and we can discuss it today in our meeting. all i know this file was meant to be in `/extras` and then decided to pull it a level up.

icaoberg: @mccalluc about the lowercase it really depends to be honest. moving forward I can suggest it as a best practice.

src/ingest_validation_tools/directory-schemas/imc3d.yaml
```diff
@@ -1,13 +1,13 @@
 # mcd/:
 -
-  pattern: 'mcd/[^/]+_HuBMAP_[^/]+organ[^/]+_slide[^/]+\.zip'
-  description: 'csv containing labels for sections as well as whether or not they were included in the 3D model'
+  pattern: 'mcd/[^/]+_HuBMAP_[^/]+_slide[^/]+\.zip'
+  description: 'CSV containing labels for sections as well as whether or not they were included in the 3D model'
```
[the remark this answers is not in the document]
mccalluc: If the zipped file is a csv, can we say the the extension should really be: `...csv.zip`? It's conventional to indicate what a `.zip` or `.gz` is once you unzip it.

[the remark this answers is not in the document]
cebriggs7135: The panel.csv should be preserved in the "extras" folder because:

- the panel.csv is provided by Zurich and is valued as an "original" source of information.
- the tag-antibody-targets in the panel.csv are ordered by the atomic weight of each metal tag which represents the order of the peaks in the mass spec data. Even though the ID info is somewhat redundant info provided in the antibodies TSV and the channelnames_report, it is considered important by the data providers.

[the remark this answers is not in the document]
mccalluc: @cebriggs7135 :
- If this would be useful for future users, put it in the PR. Information that's just in the discussion here will be hard for them to find.

@icaoberg : 
- Chris says this should be be in extras -- please update the PR if that is correct. 
- At a minimum, what are the columns the the CSV?
- If we're still in a phase where we can nudge the TMC, please ask for the file to end `.csv.zip`... but if it's too late for that, please say so in the PR and I won't nag.

src/ingest_validation_tools/directory-schemas/imc3d.yaml
```diff
@@ -1,13 +1,13 @@
...
   required: True
   is_qa_qc: False
 
 -
   pattern: 'mcd/section_report\.csv'
-  description: 'csv containing labels for sections as well as whether or not they were included in the 3D model'
+  description: 'CSV containing labels for sections as well as whether or not they were included in the 3D model'
```
mccalluc, line 10: What are the columns?

icaoberg: @mccalluc updated the column info today.

src/ingest_validation_tools/directory-schemas/imc3d.yaml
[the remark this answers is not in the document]
icaoberg: @mccalluc 

1. That is correct. `panels.csv` is now in `extras`.
2. Descriptions have been updated.
3. Too late, but we should add this recommendation to the guide. However, in terms of the reg expression, can we do `.zip|.csv.zip`? Would that be an acceptable practice?

jswelling (approved):

PR #613, opened 2021-03-10
Simple update to pattern in CODEX directory structure

icaoberg:
Validation was not working because the reg expression did not match tifs starting with the string `NAV`.


mccalluc (approved):
Looks good: I've regenerated the docs and pushed an update.

Since you've flagged Joel as a reviewer, I'll wait for his approval. (If you hadn't, I would feel comfortable just going ahead with the merge.)

jswelling (commented):
So according to the directory schema this is Akoya only.  Have we made a decision to exclude the Stanford format, or is there to be another update when we have Stanford data on line?

icaoberg:
> So according to the directory schema this is Akoya only. Have we made a decision to exclude the Stanford format, or is there to be another update when we have Stanford data on line?

@jswelling that is correct. To the best of knowledge we have excluded Stanford for the March goals. @pdblood is working on reaching out to Stanford.

This update is simply a typo fix, not an update on the schema based on Stanford.

PR #614, opened 2021-03-12
More thoughts on directory structure

mccalluc:
In the future, I would be happy to be out of the loop on the directory schemas.

Until then, one thing that could help me in the PRs is to understand whether bad naming decisions were made months ago, and we just need to live with them... or if the TMCs would still be responsive to feedback that would save future users some grief. 

cc @icaoberg , @pdblood


cebriggs7135:
@mccalluc @pdblood @icaoberg I don't foresee any TMCs being resistance to new naming rules. We're still early in the game. As long as the rules are clearly articulated to data submitters ahead of time.

cebriggs7135 (approved):

PR #628, opened 2021-03-19
scrnaseq yaml modified for external metadata

cebriggs7135:
"required" status changed across several fields to accommodate HCA 10x metadata.
One typo fixed and description text in "is_targeted" edited.


mccalluc (commented):
As it says at the top of the file:
```
```
but I'll copy this over, and add `source_project` as the first column. 

If the typo corrections matter, please make them in the original as well.

PR #631, opened 2021-03-22
scrnaseq yaml for external (non-HuBMAP) data

cebriggs7135:
"Required" status changed to false for 11 fields.


mccalluc (changes requested):
No. 
- HCA has a separate schema: https://github.com/hubmapconsortium/ingest-validation-tools/blob/master/src/ingest_validation_tools/table-schemas/assays/scrnaseq-hca-v0.yaml ... Any changes for HCA need to be made there ... We must be clear about the distinction between the HuBMAP and HCA schemas!
- Please do not add `required: true`: it is the default, and stating it explicitly just adds clutter.

I'll close this PR to avoid confusion.

PR #634, opened 2021-03-22
Support python 3.6

### LINKED ISSUE #630: Downgrade python to 3.6
Joel writes:
>  yes, unfortunately PROD, DEV &etc are at 3.6 because that's the version we can get from CentOS rhl. 

- Downgrade in Travis.
- Replace dataclass with namedtuple.
- ...?

mccalluc:
Fix #630


jswelling (approved):
Thanks for this.  The need for 3.6 is a growing annoyance but we are not yet ready to abandon it.

PR #635, opened 2021-03-22
assays version 1

### LINKED ISSUE #537: Version all the things!
- Tag version 0 of everything
- Add version columns to everything. 
- Add a script for version bumping

### LINKED ISSUE #548: Add a "description" field to all metadata schemas
- Does this change have buy-in from other groups providing the same kind of data? Haven't asked
- Can you quote or reference any outside documentation that motivates this change? No.
- Is this change backward compatible, or does it make validation more stringent? Descriptions were asked for previously, this just is a different channel.

Slack context: 
Chuck:
> In the document Chris and Alex are preparing, there’s some confusion about dataset descriptions … since the datasets are no longer created individually in ingest-ui, it will not be feasible for the TMCs to provide these descriptions in the ingest-ui, and it’s not something we’ve articulated another channel for.
>
> My sense is that the descriptions supplied by the TMCs have been of varying quality, and even under the best circumstances, it’s a minimal value add over what can be constructed automatically, for example. (If there is something really important that is only going in the description, perhaps think about getting it into the metadata?)

Bill:
> We discussed two levels of “description”- one that we can autogenerate and a second “why” description that will need to be collected in the tsv.  Can we collect this second version in the tsv?

(@cebriggs7135 : If this goes forward, please make one tweak to the level-1.yaml, rather than editing every assay type individually.)

@cebriggs7135 will evaluate this request, and if it can move forward,
will file a PR for @ngehlenborg to review,
and @mccalluc will make sure tests pass before merging.

### LINKED ISSUE #551: New IMC version with antibodies_path field
Please add an antibodies_path field.

### LINKED ISSUE #561: IMC Change ... metadata schema: ...
The IMC schema includes 2 datetime fields that are not necessary. The execution_datetime field captures the start_datetime and the end_datetime is not relevant. Please remove these 2 fields:

start_datetime and end_datetime

Per 2/19/21 conversation with the IMC team from UFLA/Zurich:

- Does this change have buy-in from other groups providing the same kind of data?
- Can you quote or reference any outside documentation that motivates this change?
- Is this change backward compatible, or does it make validation more stringent?

@cebriggs7135 will evaluate this request, and if it can move forward,
will file a PR for @ngehlenborg to review,
and @mccalluc will make sure tests pass before merging.

### LINKED ISSUE #604: Add `is_contact` to contributors schema
Required boolean field. Necessary for minting DOIs.

@shirey : Will this do what you need?

mccalluc:
Fix #551 / Fix #561 / Fix #604 / Fix #537 / Fix #548

@cebriggs7135 / @jswelling -- Heads up: PR is huge, so don't feel obliged to scroll through it.


mccalluc:
I want to add some tests before merging ... I want to make sure we maintain compatibility with both the old and new versions.

PR #636, opened 2021-03-22
"Required" field status changes for HCA

cebriggs7135:
18 fields are Required = False.


mccalluc:
I only see one field that you've made optional. Should there be others?

mccalluc (commented):
> 18 fields are Required = False.

I only see you changing two fields... Everything else is the way it should be?

cebriggs7135:
Only 2 changes showed up here. Should be 18.

PR #640, opened 2021-03-23
Mccalluc/empty row col cleanup

### LINKED ISSUE #633: add empty row/col removal to cleanup_whitespace.py

mccalluc:
Fix #633.

This script is tested two different ways:
- We generate and then run it against test examples with various encodings.
- Internally, there are doc tests, which are better for checking logic, but it would be a pain to try put character encodings in them.

Besides a refactoring, the only change here is that blank columns and rows are now stripped, and demonstrated in the doctest:
```
    >>> clean([
    ...     ['  x', 'y  ', ''],
    ...     ['', '  Hi!  ', '', ''],
    ...     ['', ''],
    ...     []
    ... ])
    [['x', 'y'], ['', 'Hi!']]
```

I think that covers it?


cebriggs7135 (approved):

cebriggs7135:
@mccalluc Blank columns are stripped? So that doesn't result in fields to the right being moved out of alignment with rows above & below?

mccalluc:
@cebriggs7135 , please remove your approval if you have misgivings.

> Blank columns are stripped? So that doesn't result in fields to the right being moved out of alignment with rows above & below?

"Blank columns" = columns that are completely empty, ie, they don't even have a header. From the example:
```
    >>> clean([
    ...     ['  x', 'y  ', ''],
    ...     ['', '  Hi!  ', '', ''],
    ...     ['', ''],
    ...     []
    ... ])
    [['x', 'y'], ['', 'Hi!']]
```
It sees that the last column header is column 2, with "y", so in the data row below, it only returns the first two columns.

We have seen TSVs come with with empty columns to the right, and empty rows below, and this cleans up both kinds of problems.

jswelling (approved):

PR #643, opened 2021-03-23
Mccalluc/apply sequence limit

### LINKED ISSUE #586: Apply sequence limit check
Joel and Chris both think it would be fine to error after a single increment. Probably apply it to every single field by default? Follow-up to #568.

mccalluc:
Fix #586. Validation error if a field looks like it's been autoincremented by Excel, even once.
- This `custom_constraint` is applied in `schema_loader.py` to every field in every schema.
- There were a number of errors in the original implementation which were uncovered trying to actually use it.
- The auto-generated yamls can be ignored, but at the bottom of the files list there are source code changes worth a review, @jswelling.
- If a different error message would be better, let me know.


cebriggs7135 (approved):

jswelling (commented):
So is the sequence_limit always 2?  I think that's the function of the code in schema_loader, and I think that matches earlier discussion, but I would expect problems with dates.  Sometimes processing will span midnight, for example.

mccalluc:
Good catch ... I think it might be simplest to loosen this to 3 across the board, rather than special casing it?

mccalluc:
> The only date I can actually find is execution_datetime, and that's actually a full timestamp, so maybe it's a non-issue.  Is there an explicit date type?  If we can't think of any fields which are actually bare dates, I'll withdraw my comment and we can just leave the limit at 2.

PR #647, opened 2021-03-24
use "units for"

mccalluc:


cebriggs7135 (approved):

src/ingest_validation_tools/schema_loader.py
```diff
@@ -207,6 +209,17 @@ def _add_constraints(field, optional_fields, offline=None):
...
         field['format'] = 'email'
+    if field['name'].endswith('_unit'):
+        # Issues have been filed to make names more consistent:
+        # https://github.com/hubmapconsortium/ingest-validation-tools/issues/645
+        # https://github.com/hubmapconsortium/ingest-validation-tools/issues/646
+        for suffix in ['_value', '']:
+            target = field['name'].replace('_unit', suffix)
+            if target in names:
+                break
+        if target in names:
```
jswelling, line 220: What happens here if suffix was not in _value or ''?  Does target not have a value when the test occurs?

mccalluc: @jswelling -- If it can't find the corresponding value field it does nothing. Another approach would be:
```python
assert target in names, f"Can't find a target for {field['name']} in {names}"
```

I don't have a strong preference: Should I make it stricter?

jswelling: Ignore this; I figured it out.

jswelling (approved):

PR #651, opened 2021-03-24
Mccalluc/consistent sequencing read format

### LINKED ISSUE #619: Consistent validation for sequencing_read_format
- Apply consistent validation to sequencing_read_format.
- Add an example in the docs if people have trouble reading the regex.

_Originally posted by @mccalluc in https://github.com/hubmapconsortium/ingest-validation-tools/issues/607#issuecomment-800605289_

mccalluc:
Fix #619. @cebriggs7135 -- Is there any compelling reason for the two assays to use different delimiters for this field? Is there any reason in the new v1 schema we can't be consistent? 

(Arbitrarily picking `/` as the delimiter ... `+` would be fine, too.)


cebriggs7135:
No reason there can't be consistent going forward.

cebriggs7135 (approved):

PR #652, opened 2021-03-24
Add missing constraints to scrnaseq

### LINKED ISSUE #590: scrnaseq.yaml is missing constraints
`diff src/ingest_validation_tools/table-schemas/level-2/{scatacseq.yaml,scrnaseq.yaml} -y` shows lots of fields which could be aligned better.

mccalluc:
Fix #590. I'm guessing at some point I was just given a list of fields, and we missed all the constraints that should be applied. If any of the constraints for scrnaseq should be different from atacseq, please suggest a correction in `scrnaseq-v1.yaml`.


docs/scrnaseq/README.md
```diff
@@ -209,6 +210,7 @@ The method by which specific cell populations are sorted or enriched. Leave blan
 | constraint | value |
 | --- | --- |
 | required | `False` |
```
mccalluc, line 212: @cebriggs7135 : Does it still make sense for this field to be optional, or should they just supply the string `none`, to be consistent with atacseq?

cebriggs7135: Might as well keep it consistent with atacseq by allowing "none".

mccalluc:
@cebriggs7135 : `none` is allowed, and the field is required. Please approve, if it should be approved.

cebriggs7135 (approved):

PR #653, opened 2021-03-24
Lightsheet directory schema

icaoberg:
First version of the Lightsheet directory schema.

Test: this schema was used to validate a dataset from UFL.


src/ingest_validation_tools/directory-schemas/lightsheet.yaml
```diff
@@ -1,3 +1,72 @@
+# Level0/
 -
-  pattern: 'TODO'
-  description: 'Directory structure not yet specified.'
+  pattern: 'Level0/Channel+[^/]/[^/]+\.csv'
```
mccalluc, line 3:
```suggestion
  pattern: 'Level0/Channel[^/]+/[^/]+\.csv'
```

What you have allows the `l` to be repeated.

mccalluc (changes requested):
Check the regex syntax. I remain concerned when there are CSVs and we don't say what the columns are. Tests need to pass before merge, and ideally before review is requested.

... but zooming out, I think I'm only on this PR because I'm a codeowner? Is the right course for me to remove myself, so I won't be auto-assigned in the future?

icaoberg:
> Check the regex syntax. I remain concerned when there are CSVs and we don't say what the columns are. Tests need to pass before merge, and ideally before review is requested.
> 
> ... but zooming out, I think I'm only on this PR because I'm a codeowner? Is the right course for me to remove myself, so I won't be auto-assigned in the future?

Agree. I was able to get some information from University of Florida. I can certainly ask them to revise the description.

jswelling (approved):

jswelling:
@icaoberg I've reviewed and approved, but I'm not sure who is supposed to actually resolve the conflicted files and do the merge.  @mccalluc does it fall to @icaoberg to carry out the actual merge?

mccalluc:
Yes, @icaoberg can get tests to pass and then merge.

jswelling (commented):
So basically the only thing required is the unstitched ome.tiff?  Are there combinations of not-required files that we expect to see together? I know we don't have a way to enforce this now, but only requiring the one file seems pretty sparse.
Also, do we need to validate the integrity of the .czi file?

icaoberg:
> So basically the only thing required is the unstitched ome.tiff? Are there combinations of not-required files that we expect to see together? I know we don't have a way to enforce this now, but only requiring the one file seems pretty sparse.
> Also, do we need to validate the integrity of the .czi file?

@jswelling to the best of my knowledge this is correct. Hunter from Florida already did a preliminary local validation. 

I am not aware if we need to validate the czi, . Most DP are providing the .czi because this is the output file from their microscopes/devices.

PR #661, opened 2021-04-02
distinguish v2 and v3 10x

mccalluc:
Following up from the conversion at 10, and I think this is one of the next steps. 
- Note that I am only touching the v1 schema... so if there are still TSVs without a `version` column, they would not allow the new values. Is that the right call?
- Also, I don't think this change merits an entirely new version of the metadata schema, but I could be mistaken: With this change a TSV that validated with the v1 schema today would not validate after the merge.

(There's an outstanding task to make sure that the terms used here align with Commons.)


cebriggs7135:
I agree that it doesn't require a new version. Not sure if it matters that metadata in the portal in the original form not designating 10X version.

jswelling (approved):
Works for me.  We'd need a matching change to assay_types.yml I presume.

cebriggs7135 (approved):

mruffalo (approved):

PR #668, opened 2021-04-07
Mccalluc/expected cell count

### LINKED ISSUE #667: Add optional expected cell counts to rna seq
@mruffalo : `expected_cell_count`? Did you suggest this should be applied to multiple types?

@mruffalo / @cebriggs7135 : description for this field? any preferences about where it goes in the list of fields?

mccalluc:
Fix  #667

- Is this the right assay? Should it also be applied to bulkrnaseq?
- Is the field description adequate?
- Should the description for the assay describe what changed with each version? ... or if someone really cares about that, can they just use the git history?
- Do we want to set a timeline for deprecating the previous version?
- While Airflow needs to continue to allow all versions, do we want to steer the TMCs to submit new data with new versions? I could imagine a flag that requires validity against the most recent version of a schema, and not just any version.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.10 - in progress
+- Add expected_cells_count.
```
mruffalo, line 4: The field name is `expected_cell_count` (singular).

docs/field-descriptions.yaml
```diff
@@ -77,6 +77,7 @@ execution_datetime: Start date and time of assay, typically a date-time stamped
   generated by the acquisition instrument. YYYY-MM-DD hh:mm, where YYYY is the year,
   MM is the month with leading 0s, and DD is the day with leading 0s, hh is the hour
   with leading zeros, mm are the minutes with leading zeros.
+expected_cell_count: Parameter for downstream pipeline; How many cells are expected?
```
mruffalo, line 80: Might be better as "How many cells are expected? This may be used in downstream pipelines to guide selection of cell barcodes or segmentation parameters."

I.e., we'll use this in the Salmon s*RNA-seq pipeline very soon, but not yet scATAC-seq or any imaging modalities, but that may change in the future.

mruffalo (changes requested):
Minor rewording and fixing the field name, but otherwise looks good!

docs/field-descriptions.yaml
mccalluc: fixed. Thanks!

CHANGELOG.md
mccalluc: fixed. Thanks!

jswelling (changes requested):
Is it correct that expected_cell_count is not required in assays/scrnaseq-v2.yaml but *is* required in docs/scrnaseq/v2.yaml?

mccalluc:
> Is it correct that expected_cell_count is not required in assays/scrnaseq-v2.yaml but _is_ required in docs/scrnaseq/v2.yaml?

Yes: [We add a number of contraints](https://github.com/hubmapconsortium/ingest-validation-tools/blob/126b207a001dc4f4d3698a0c0052a36e6ef48222/src/ingest_validation_tools/schema_loader.py#L164) to the base schemas in `table-schemas/assays`; Most are based on the name of the field, but we also make fields required, unless the schema explicitly says `required: false`. (This is the opposite of the default behavior of the tableschema library, which assumed fields are optional unless marked as required.)

The schemas under `docs/` are a record of the result of transforming the schemas in `src/`. Their primary purpose is to alert us if new code causes unexpected changed to the actual schemas: They can be read as a description of what the validation rules _really_ are, with nothing being implicit.

mccalluc:
@jswelling / @mruffalo : I'll re-request review from you two.

On our side, we're been using "comment" to mean "here's something to fix, but I trust you to finish it" and "request changes" as "after fixes or explanation, I want to review it again". If that model doesn't work for you, let me know.

jswelling:
Remind me what happens for a field that is 'required' but empty- that violates the schema, does it not?  I think we will have submissions for which the expected cell count is unknown, and I think it is sensible to leave the field blank in that case.

mccalluc:
@jswelling : it is optional:
```
    name: expected_cell_count
    description: How many cells are expected? This may be used in downstream pipelines to guide selection of cell barcodes or segmentation parameters.
    type: integer
    constraints:
      required: false
```

If there's something in here which suggests otherwise, please point it out to me?

jswelling (approved):

cebriggs7135 (approved):

mccalluc:
Addressed @mruffalo's comments earlier... going ahead with merge... but let me know if you'd like me to wait in the future.

PR #670, opened 2021-04-09
Remove the sequence_limit where not appropriate

### LINKED ISSUE #669: Turn off auto-increment check for antibodies
Antibodies `channel_id` is expected to be sequential integers!

mccalluc:
The constraints are really applied on a field-by-field basis... I could list all the antibodies field, but I'm not sure that's called for... I know that the `channel_id` is often sequential -- Are there others to add? Fix #669.


mccalluc:
> Only the channel ID field has increments.

@cebriggs7135  -- That is what this PR does. Please formally approve if that is your intention.

jswelling (approved):
That should do it.  I'm honestly unsure if we'll encounter other fields that tend to include sequential values, but we can take them on a case-by-case basis.

cebriggs7135 (approved):

Issue #663, opened 2021-04-02
Maps in the YAML for ontology backed enums


mccalluc:
We're going to go with an approach where the TSVs contain human-readable strings, and these are expanded to ontology URIs (from a predefined list) in a later phase of ingest.

For the test run, I would suggest first doing this with the Donor YAML. It's not actually used for anything, so tweaks there shouldn't cause anyone any trouble, and we won't have tons of merge conflicts. (Why do we even have the Donor YAML? Even though Donor metadata has an entirely different ingest route, I think Jonathan felt that for the sake of documentation, it should be included here.)

What needs to be done? (*Italics* may need more attention, clarification.)
- For enums, which currently are lists, maps will be allowed, with the keys being the human-readable strings, and the values the URIs. (Normally, I'd say the URIs would be better keys, but in this case, I'm mostly concerned about keeping the yaml readable, so I want the human readable on the left.)
- When one of these mapped enums is encountered during doc generation, the docs will be more informative... perhaps there's a separate table of term/URI pairs. *Let me know if there are more requirements here.*
- When one of these mapped enums is encountered during template generation, the pull-down will have an extra value, and the sheet will have an extra, optional, column next to the enumerated column.
- *This none-of-the-above options will still cause a validation error: Chuck would really prefer to have a shared understanding with the TMC that if they are suggesting new terms, some errors will be expected.*
- *We do not forbid users from selecting a value, and entering a value in the adjoining free text field... They really shouldn't, but maybe they want to suggest a refinement on the main term? In any case, adding conditional validation on the free-text field is more complexity I'd like to avoid. Sorting this out is a curation responsibility.*
- Finally, there is a standalone script that complements the column with terms by adding a column with the URIs. This will be incorporated, somehow, into Joel's workflow, and will run after the last validation. (This expanded schema will not validate, because of the additional columns, but that seems ok.)
- Finally, finally, update the process diagram on the readme to point at https://docs.google.com/drawings/d/11DLkPe_mfC3kh69HCYqCx5MAVTTkYOXqVjyFY85jMLQ/edit

cebriggs7135:
@mccalluc Can you explain this sentence in further detail because I'm not sure what "doc generation" means or refers to?

*When one of these mapped enums is encountered during doc generation, the docs will be more informative... perhaps there's a separate table of term/URI pairs. Let me know if there are more requirements here.*

mccalluc:
[...]

PR #672, opened 2021-04-12
Mccalluc/ontology enums

mccalluc:
Towards #663.

- @ngehlenborg / @cebriggs7135 -- Basis for discussion in the next UI meeting. Questions
  - Is the second table in the markdown good?
  - What precisely should the additional pull-down value be? (In this PR, it is currently `None of the above`.)
  - What precisely should the additional column be called, and what should the description be? (Currently: `blood_type_desired`, `Desired value for ____`)

- @jswelling -- Are you on-board with the idea of a post-validation step to be run by airflow, to add new columns to the TSVs with ontology URIs, based on a mapping supplied in the schemas?
  - Is a function which takes a TSV path and returns the expanded TSV as a string the best thing to deliver to you?
  - Are you Ok with deploying new versions of this code to Airflow as Chris identifies new terms to add? (Hopefully, not frequent, but it will happen.)

- @ilan-gold  -- This cuts across many areas of the code, so it might be a good entry point to the rest of the repo. Does it make some sense?

Next steps: 
- I would like to get some feedback now, but before merging, 
- I would like to get the real URI mappings for some assay, and introduce a new version of that assay with the mappings,
- and make sure validation does not allow "None of the above",
- and then I would like to revert the changes to Donor,
- and when we have an assay with ontology terms, finish `expand_tsv.py` and make sure it works,
- and finally, remove the orange circles from the updated diagram.

### Updated submission process with ontology enums
[Submission process](https://docs.google.com/drawings/d/11DLkPe_mfC3kh69HCYqCx5MAVTTkYOXqVjyFY85jMLQ/edit)

### Even more updated, with enums, and my guess about Bill's Upload work.
[Submission process](https://docs.google.com/drawings/d/1k1h_sPWo-oD5eMw7wGcBn4cuN4eZF6P_dAAyo5CnGIE/edit)


cebriggs7135:
@mccalluc Please clarify "second table in the markdown".

README.md
```diff
@@ -147,4 +147,4 @@ but if they are applicable to only a single dataset, they can be placed within t
 
 You can validate your submission directory locally, then upload it to Globus, and the same validation will be run there.
 
-[![Submission diagram](https://docs.google.com/drawings/d/e/2PACX-1vS8F78bk0zHSRygMIyTLruAMxjL4c5EY_q_Mp3gN2TbdZLtalax5AxyvwBWyqWwAJH941ziqJPqBDTW/pub?w=500)](https://docs.google.com/drawings/d/1nhrRWBgcZh6GE2MCKysIq4KzsRL6SZm0jYtvadF83Kk/edit)
+[![Submission process](https://docs.google.com/drawings/d/e/2PACX-1vScOVTJCQjaHj0oiVuvPxuQDNXFCtI-CGTTOm-kSPji9DehMxRIaWg98qKAkkvwSWn5d2TuL1z4mIyh/pub?w=1000)](https://docs.google.com/drawings/d/11DLkPe_mfC3kh69HCYqCx5MAVTTkYOXqVjyFY85jMLQ/edit)
```
cebriggs7135, line 150: Chuck, this is the wrong link.

cebriggs7135 (approved):

docs/donor/README.md
```diff
@@ -0,0 +1,115 @@
...
+| constraint | value |
+| --- | --- |
+| enum | `A`, `B`, `AB`, or `O` |
+| required | `True` |
+
+Ontology terms:
+
+| term | URI |
+| --- | --- |
+| A | `http://purl.bioontology.org/ontology/SNOMEDCT/49790006` |
```
ngehlenborg, line 52: I am not sure if we should include the definitions of the terms here or not. It would save submitters one click but would be a lot of work to maintain this if we include it here. 

What was your rationale for not including a definition?

docs/donor/donor.tsv
```diff
@@ -0,0 +1,2 @@
+age_value	blood_type	blood_type_desired	body_mass_index_value	cause_of_death	height_unit	kidney_donor_profile_index_value	mechanism_of_injury	medical_history	race	sex	weight_value
```
ngehlenborg, line 1:
```suggestion
age_value	blood_type	blood_type_suggested	body_mass_index_value	cause_of_death	height_unit	kidney_donor_profile_index_value	mechanism_of_injury	medical_history	race	sex	weight_value
```

docs/donor/donor.tsv
```diff
@@ -0,0 +1,2 @@
...
+	A / B / AB / O / None of the above
```
ngehlenborg, line 2:
```suggestion
	A / B / AB / O / Submitter Suggestion										
```

ngehlenborg (changes requested):
I am unsure about including definitions in the ontology table or not. But I would be willing to accept the current solution.

Instead of "None of the Above" and "_desired" I am proposing "Submitter Suggestion" and "_suggested", respectively.

src/ingest_validation_tools/docs_utils.py
```diff
@@ -170,7 +170,20 @@ def _make_constraints_table(field):
...
         table_md_rows = []
-    return '\n'.join(table_md_rows)
+    main_table_md = '\n'.join(table_md_rows)
+
+    onto_table_md = _make_onto_table(field['constraints']['enum']) \
+        if 'constraints' in field and 'enum' in field['constraints'] else ''
+    return main_table_md + onto_table_md
+
+
+def _make_onto_table(enum):
```
ilan-gold, line 180: maybe spell `ontology` - it almost reads like a typo into -> onto

ilan-gold (approved):
Looks good.  Thanks for looping me in.

jswelling (commented):
@mccalluc could you give me an example of a transformation which might be performed by Airflow on a .tsv as a result of this PR?  I'm unclear on why any updates get written back to the .tsv rather than simply updating neo4j.  Is the notion to allow the user to later download the .tsv and have it include ontology updates that happened after the dataset was ingested?  At what point in the workflow would the write-back occur- before multi-dataset uploads are restructured, or after?

mccalluc:
@jswelling :
>  I'm unclear on why any updates get written back to the .tsv rather than simply updating neo4j.

You're right -- trying to update the TSV introduces unnecessary problems. Should I deliver just a module to you that expands the terms, filling in the corresponding URIs from the schema, or would something else be better?

jswelling:
I'm still behind the curve. Can you give me an example of a pair in the original dict and the URI that would be added in the returned dict?
I think it's very reasonable to try to elaborate the metadata dict between the time it is read from the metadata.tsv files and the time it is passed to ingest-api .  I'm just unclear on what the elaborations would be and why they would be coming from ingest-validation-tools.  Maybe we need an ingest-elaboration-tools?

mccalluc:
@jswelling 
> Can you give me an example of a pair in the original dict and the URI that would be added in the returned dict?

Unfortunately, I don't have any real examples of assay metadata fields that should be based on ontologies, along with their actually ontology IDs. I wish I did. I've tweaked `blood_type` on Donor to have something to work with, but I would revert that before merging.

... so, an entirely made up example: ice cream. If the src schema defines:
```yaml
fields:
  - name: flavor
    constraints:
      enum:
        chocolate: http://example.com/choco
        vanilla: http://example.com/vanil
```
the validation code expands that to:
```yaml
fields:
  - name: flavor
    type: string
    constraints:
      required: true
      enum:
        chocolate: http://example.com/choco
        vanilla: http://example.com/vanil
  - name: flavor_suggested
    type: string
    required: false
```
and if a TSV like this comes in:

| flavor | flavor_suggested |
| --- | --- |
| None of the above | strawbery |

Then:
- Curation will identify an appropriate ontology term, perhaps with help from an ontology specialist designated by Jonathan.
- Curation will start a PR, adding that term to the enum.
- A developer will complete the PR and merge to master.
- IEC will need to update the ingest-validation-tools that is run by airflow.
- Curation will need to update the TSV on Globus and re-run airflow.

_(The last two steps concern me: Is getting new code out deployed going to be hard on your side? Is curation going to be able to make edits to the TSV?)_

After all of that, the src schema will look like:
```yaml
fields:
  - name: flavor
    constraints:
      enum:
        chocolate: http://example.com/choco
        vanilla: http://example.com/vanil
        strawberry: http://example.com/straw
```
and the TSV will have the value suggested by the user (in this case misspelled), along with the ontology term determined by curation:
| flavor | flavor_suggested |
| --- | --- |
| strawberry | strawbery |

Now, validation runs smoothly, and the dict that you get by reading this row:
```json
{
  "flavor": "strawberry",
  "flavor_suggested": "strawbery"
}
```
needs to be expanded to:
```
{
  "flavor": "strawberry",
  "flavor_suggested": "strawbery",
  "flavor_uri": "http://example.com/straw"
}
```
before it goes into Neo4J and on to Elasticsearch.

Actually, I want to rewind just that last bit: The important part is that the terms which are based on ontologies are actually in the knowledge graph -- I don't think it's that important to have another string field, in this case a URI -- what we really need is for `http://example.com/straw` to be in the graph, so if `strawberry` is a hyponym of `fruit`, we'd be able to locate this record by searching for the higher level term.

docs/donor/donor.tsv
mccalluc: fixed.

docs/donor/donor.tsv
mccalluc: fixed.

docs/donor/README.md
mccalluc: Summarizing the discussion yesterday:

- We'd like to utilize IEC's knowledge graph here: Either they provide an API we can hit at doc generation time to get a definition for a term, or, simpler, we just generate a link parameterized by the URI to a human-readable page supported by the HuBMAP ontology service.
- We **should not** link to the uri, or attempt to use HTTP on it.
  - wikipedia: ["Other URIs provide only a unique name, without a means of locating or retrieving the resource or information about it"](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)
  - w3c: ["The argument that `http:` URIs are "locations" is based upon incomplete understanding of the use of URIs."](https://www.w3.org/2001/tag/doc/URNsAndRegistries-50.html#iddiv2137969152)
  - SO: ["in general all URI cannot be referenced"](https://stackoverflow.com/questions/1115503/use-of-uri-and-url-in-semantic-web)

src/ingest_validation_tools/docs_utils.py
mccalluc: fixed

README.md
mccalluc: fixed.

jswelling:
Ah, great example.  So the original .tsv file contains "strawbery" in the "flavor" column.  This fails validation.  Does the validator query the ontology at validation time, or is the metadata schema built automatically from the ontology elements?  How did "chocolate" and "vanilla" get into the schema in the first place- did the schema learn of them from the ontology or the other way around?
Also, what does the validation code do with its expanded version- is that presented to the submitter as an error?  At the moment there is no route by which the information in that bit of yaml can make it to the curator, who is probably different from the submitter.  Curation comes after submission, which comes after clean validation, so how does the curator know to add 'strawberry' to the ontology or the schema?  How does the curator let the submitter know that they're good to validate if they replace "strawbery" with "strawberry"?

mccalluc:
@jswelling 

> So the original .tsv file contains "strawbery" in the "flavor" column. This fails validation. Does the validator query the ontology at validation time, or is the metadata schema built automatically from the ontology elements?

Neither, if by "query the ontology" you mean make a request to the ontology service. The validation of terms is based on the values in the schema. In the current proposal there is not process for ontology terms akin to the network checks.

> How did "chocolate" and "vanilla" get into the schema in the first place- did the schema learn of them from the ontology or the other way around?

My sense is that Jonathan's knowledge graph will contain every term we'd want, so selections would be made from it to fill the enum in the schema... In the event that adding a new ontology term is warranted, that will need to be both in the knowledge graph and in the schema: Neither one feeds into the other.

> Also, what does the validation code do with its expanded version- is that presented to the submitter as an error?

I'm not quite sure what you mean by "expanded version" ... the URI for a term? but I'm not sure why that would be an error, so I may be misunderstanding. Please clarify.

> Curation comes after submission, which comes after clean validation, so how does the curator know to add 'strawberry' to the ontology or the schema? How does the curator let the submitter know that they're good to validate if they replace "strawbery" with "strawberry"?

We are loosening the assumption that curation comes after a clean validation:

Screen Shot 2021-04-14 at 6 31 17 PM

If a TMC needs a term that isn't yet in the schema, they will fill in what they want ("strawbery" in this case), and this is a validation error they are allowed to have at submission time. (Note: It's a validation error not because "strawbery" is misspelled, but because it is not in the schema.) The curator would find the correct term to match the submitter's intent, update the TSV, and the schema, and somehow get those changes deployed on airflow, and revalidate. (This is smaller loop in the flowchart above.)

Honestly, this does worry me: We talked about having the TMCs communicate with us before submission, so we would get new terms in before they validate, but I think for Nils that involves too much communication back and forth -- He would rather they just submit, and then it's in our court.

jswelling:
> I'm not quite sure what you mean by "expanded version" ... the URI for a term? but I'm not sure why that would be an error, so I may be misunderstanding. Please clarify

I'm referring to the block of yaml which follows "the validation code expands that to" in your original discussion.  What does that piece of yaml get used for?

jswelling:
> Honestly, this does worry me: We talked about having the TMCs communicate with us before submission, so we would get new terms in before they validate, but I think for Nils that involves too much communication back and forth -- He would rather they just submit, and then it's in our court.

Yes, this worries me too.  The validation process is already pretty fraught, and adding this kind of semantic negotiation sounds really impractical to me.  I'm also concerned about the fact that terms are redundantly defined in the schema and in the ontology- that violates the principle of localization.  It is the ontology's job to know that 'strawberry' is a flavor.  If the validator wants to know that a certain string represents a flavor, it should ask the ontology.

mccalluc:
@jswelling :
>  What does that piece of yaml get used for?
```
fields:
  - name: flavor
    type: string
    constraints:
      required: true
      enum:
        chocolate: http://example.com/choco
        vanilla: http://example.com/vanil
  - name: flavor_suggested
    type: string
    required: false
```
- It is used for generating the Excel file, so for every enumerated field that is linked to an ontology, there's a corresponding `_suggested` field.
- and it is used for validation of the TSVs, since they will have this pair of columns

> The validation process is already pretty fraught, and adding this kind of semantic negotiation sounds really impractical to me.

The design is influenced by the practices of other consortia, where the data providers aren't blocked if the terms they want aren't available: they fill out the metadata as best they can, suggest the terms they need if they can't find approved terms, and finding the _correct_ term is a task left for curation.

> I'm also concerned about the fact that terms are redundantly defined in the schema and in the ontology- that violates the principle of localization. It is the ontology's job to know that 'strawberry' is a flavor. If the validator wants to know that a certain string represents a flavor, it should ask the ontology.

Two things to note:
- Human-readable terms can't consistently be resolved to unique URIs. 
- The set of allowed terms is something curation needs to be involved in, even when they are backed by an ontology.

"It is the ontology's job to know that 'strawberry' is a flavor" -- I don't think that's really true. Given "white", it might be a color, or an ethnicity, or part of the anatomy of the eye.

Perhaps we could say: "find 'white' where it is an anatomy term" ... even then, we aren't guaranteed uniqueness, unless the curator is somehow involved in updating the knowledge graph to identify sets of terms which exclude homonyms. More generally, I think we have an assumption that adding terms is something that should take some curator involvement.

Stepping back, it's good to acknowledge that I do have a bias against relying on the ontology service at run time. For one, does it even exist? ["Ontology API" here](https://portal.hubmapconsortium.org/docs/apis) points to the CCF docs, and I don't see an API endpoint. But beyond that: validating against a fixed set of terms is something we do all the time: Constructing queries to validate against a service is entirely new ... While I'm sure it's possible, it is more complexity, and I want to be sure if we go there, it's because the user stories require it, not because we think we should use an API hammer for this nail.

mccalluc:
@ngehlenborg / @jswelling :
- Updated diagram: Changed "Submission" to "Upload", and changed the wording in the boxes that were circled.
- Removed the proof-of-concept `expand_tsv.py` script.

Again, blood types are just a demo: Nothing uses the Donor schema, and as soon as we have some ontology terms where we actually need them, in the assay metadata, the blood-type demo could be removed.

It sounded like we were on the same page, but if there are focused changes necessary, great: Please point them out.

docs/donor/README.md
ngehlenborg: Are there any recommendations on how to resolve ontology URIs?

ngehlenborg (approved):

jswelling (changes requested):
* Could you add a mark on the diagram showing the point where the upload passes from TMC control to IEC control?  I believe it would be just before or just after 'Curation slacks IEC the Upload UUID'.
* There is code development in progress that will allow the TMC to trigger an Airflow validation and read the validation report.  These are tasks labeled as being done by Curation in the diagram, but the hope is that in the great majority of cases Curation won't have to be involved.  How does that affect the diagram?
* I remain concerned that adding complexity to the block where validation is performed immediately after upload will slow a process that is already too slow, but I recognize that adding terms to the enums will be rare.

mccalluc:
> - Could you add a mark on the diagram showing the point where the upload passes from TMC control to IEC control? I believe it would be just before or just after 'Curation slacks IEC the Upload UUID'.

I have added a dashed line separating TMC responsibility from HIVE responsibility... but it's not a clean division: For example, up in the upper right, we say the TMC can change the data on Globus if they can't give final approval... I'd have a slight preference to leave this new line out, but I don't care strongly. 

> - There is code development in progress that will allow the TMC to trigger an Airflow validation and read the validation report. These are tasks labeled as being done by Curation in the diagram, but the hope is that in the great majority of cases Curation won't have to be involved. How does that affect the diagram?

Possible revision [here](https://docs.google.com/drawings/d/1TIWsvb8qO3TNvypOGRB2kWYq4PTBATD1zYgi113UeGU/edit) -- Until that is closer to being ready to go, I'd like to keep it out of the diagram, so it's closer to our understanding of how things should be working _now_.

> - I remain concerned that adding complexity to the block where validation is performed immediately after upload will slow a process that is already too slow, but I recognize that adding terms to the enums will be rare.

I share your concern, but the marching orders from Nils (and I believe Jonathan is on board) are that TMCs can propose new terms at submission time, rather than requiring a conversation before submission to identify terms. If I've misunderstood the requirements, and we can avoid the traffic jam, I'd be happy to hear that... but I don't think that's the case.

[Here's a revision](https://docs.google.com/drawings/d/1Vav1OVlT9CDXNX8f-caNwyfkL1WHMPRejXdIv9HjWD8/edit) that splits out all the sub-steps. With the requirements given, I don't think the complexity can be avoided, but it could be explained more clearly, if that would help?

jswelling:
> I have added a dashed line separating TMC responsibility from HIVE responsibility... but it's not a clean division: For example, up in the upper right, we say the TMC can change the data on Globus if they can't give final approval... I'd have a slight preference to leave this new line out, but I don't care strongly.

I agree that this line is in the right place, and I'd like to retain it despite the issues in the upper right.  It disambiguates who is supposed to be fixing problems that occur in the block above.
 
> > * There is code development in progress that will allow the TMC to trigger an Airflow validation and read the validation report. These are tasks labeled as being done by Curation in the diagram, but the hope is that in the great majority of cases Curation won't have to be involved. How does that affect the diagram?
> 
> Possible revision [here](https://docs.google.com/drawings/d/1TIWsvb8qO3TNvypOGRB2kWYq4PTBATD1zYgi113UeGU/edit) -- Until that is closer to being ready to go, I'd like to keep it out of the diagram, so it's closer to our understanding of how things should be working _now_.

Well, it's pretty close, but that's fair.  One feature of the updated diagram which I like is that it separates repairs to be made by the TMC from those to be made by Curation/Dev/IEC.  This is also true of the broken-out version which shows the sub-steps; I like that.  In this diagram the question of how the TMC informs IEC that an upload is ready.  I *think* the plan is that they would send email to help@hubmapconsortium.org .  If that is the case, and if there were ontology changes required, they could say so in the same email.  

> [Here's a revision](https://docs.google.com/drawings/d/1Vav1OVlT9CDXNX8f-caNwyfkL1WHMPRejXdIv9HjWD8/edit) that splits out all the sub-steps. With the requirements given, I don't think the complexity can be avoided, but it could be explained more clearly, if that would help?

I agree that the best we can hope for at this point is to separate responsibilities, and to keep Curation out of the upload submission loop wherever possible.   I like this revision, at least until the change that allows TMCs to trigger Airflow validation is ready.

docs/donor/README.md
mccalluc: Eventually, the knowledge graph API, but that doesn't exist right now.

mccalluc:
Diagram revised. Merging!

PR #677, opened 2021-04-15
optional contributors for HCA

### LINKED ISSUE #676: scrnaseq-hca reduce error messages
Both contributors_path and data_path are optional. 
I leave the contributors_path empty and I leave the true fastq.gz filenames for this project in the data_path field because that info will be needed to associate each HCA dataset with the actual HCA fastq.gz files Matt has uploaded. 

The problem is that I get this message for every row:

*In the dataset /Users/christinebriggs/dropbox/HMS/INGEST_WORK/RELEASE2_DOCS/HCA/HCA_10X_TESTRUN/TLH_2nd_wash_S3_L001_I1_001.fastq.gz referenced on row 2, the file "/Users/christinebriggs/dropbox/HMS/INGEST_WORK/RELEASE2_DOCS/HCA/HCA_10X_TESTRUN/TLH_2nd_wash_S3_L001_I1_001.fastq.gz" does not exist.
In HCA_test_Liver-metadata.tsv (as scrnaseq-hca): External: row 2, contributors /Users/christinebriggs/dropbox/HMS/INGEST_WORK/RELEASE2_DOCS/HCA/HCA_10X_TESTRUN: Expected a TSV, but found a directory.*

1) Why is the empty contributors_path fields seen as containing a directory? Can that message be omitted?
2) Can you omit the "does not exist" messages in the data_path column?

mccalluc:
Fix #676: Motivating example is HCA, but this actually improves the error reporting in another test case as well, and simplifies the code.

@cebriggs7135 , copying over my comment from the issue:
>    Can you omit the "does not exist" messages in the data_path column?

Could you just use --tsv_paths instead of --local_directory?

I also want to ask if Contributors will _ever_ be used with HCA data... if not, might consider just getting rid of the column?


ilan-gold (approved):

cebriggs7135:
@mccalluc ccalluc It would be better if I could just validate the metadata tsv since the data files won't be submitted by me. Contributors are available for some HCA projects but none of the ORCID Id's are available. We should discuss with Nils tomorrow.

cebriggs7135 (approved):

PR #692, opened 2021-04-23
Mccalluc/gh pages

### LINKED ISSUE #686: Set up github pages?
@cebriggs7135 -- Would it be any help to host the documentation on github pages? This would give us a little more control on formatting, and would avoid all the github framing which I think can be off-putting?

mccalluc:
Fix #686 

Motivation: 
- Non-devs get a little scared/confused looking at docs in github.
- If people aren't reading from the `docs` directory in github, there will be less temptation to submit PRs in `docs`.
- A little branding could help people understand that this really is part of hubmap.
- Explaining that that docs are there to support gh-pages will help to justify this somewhat weird setup with outputs being checked in.

GH Pages are currently being generated from this branch, so you can see the result at https://hubmapconsortium.github.io/ingest-validation-tools/ -- For now, I don't think we'd advertise this: If someone goes to the old README, [they'll see a pointer to the index.md](https://github.com/hubmapconsortium/ingest-validation-tools/blob/9b543abdb68b9b494e64665bcd6f75019c4230d3/docs/af/README.md). Once the jekyll pages have been styled, we could make gh-pages primary.

This isn't really idiomatic Jekyll: It would be better to use collections... as it stands, Jekyll doesn't know that all these pages are of the same class... But I don't think rearchitecting this into a proper collection, (or moving the rendering out of python and into liquid,) should be a priority.

- @cebriggs7135 - Do you think this is a worthwhile goal? Will it be ok to be in this intermediate state for a while, and we'll get all the styling in place before transitioning?
- @tsliaw - Would you be ok with doing the CSS here? We can get the colors and fonts right, and a little branding, but I don't think it would be time well spent to use all of Material UI.
- @ilan-gold  - Code review.

**Before merge** Chuck should turn off the delete-branch-after-merge setting.
**After merge** Chuck should turn it back on, and should target gh-pages generation to `master`.


ilan-gold (approved):
There seems to be an extra bullet point at the top.
Screen Shot 2021-04-26 at 9 55 46 AM

Besides that I'm just curious why not use a theme and also what you mean by "python rendering?"

mccalluc:
@ilan-gold 
> There seems to be an extra bullet point at the top.
> Screen Shot 2021-04-26 at 9 55 46 AM

Thanks!

> Besides that I'm just curious why not use a theme 

I should do that -- but on top of that, I think Nils would still like it to look more HuBMAP-y. Bill is also interested in HuBMAP styling for GH Pages, so this could be re-used.

> and also what you mean by "python rendering?"

If I had started with Jekyll, then the normal thing to do would be to put the information into the yaml frontmatter, and generate the markdown with a liquid template... but right now, there's already a lot of python code to construct the markdown that we see, and I think there is little benefit to trying to translate that into liquid.

tsliaw (approved):

ilan-gold:
> If I had started with Jekyll, then the normal thing to do would be to put the information into the yaml frontmatter, and generate the markdown with a liquid template... but right now, there's already a lot of python code to construct the markdown that we see, and I think there is little benefit to trying to translate that into liquid.

I see what you are saying now, makes sense.

cebriggs7135 (approved):

PR #699, opened 2021-05-04
Fix maldiims schema

### LINKED ISSUE #696: Error in maldiims schema
Hi,

I was testing our next data release directory yesterday and noticed that my imzML directory was failing. It looks like the pattern has a minor bug in it. It should be `'imzML/[^/]+\.imzML'` not `'imzML/[^/]+\.imzMl'`.  
The index file is usually spelled as `.imzML`. 

https://github.com/hubmapconsortium/ingest-validation-tools/blob/12265bad2fd4cc17428d7ee65fb1dff2c91dfc84/src/ingest_validation_tools/directory-schemas/maldiims.yaml#L11

lukasz-migas:
- fixes maldiims schema ( closes #696  )
- added maldiims example
- updated README.md to include the `examples/...` directory in the given examples


test.sh
```diff
@@ -12,7 +12,7 @@ CONTINUE_FROM="$1"
 
 if [[ -z $CONTINUE_FROM ]]; then
   start flake8
-  flake8 || die 'Try: autopep8 --in-place --aggressive -r .'
+  flake8 src || die 'Try: autopep8 --in-place --aggressive -r .'
```
mccalluc, line 15: This is just a performance improvement, so it doesn't need to scan other directories that don't contain python?

examples/dataset-examples/good-maldiims/README.md
```diff
@@ -0,0 +1,3 @@
+```
```
mccalluc, line 1: We haven't had end-to-end tests for all assay types, so don't feel obliged to add one here.

If this exercises an execution path that isn't sufficiently tested, great, but otherwise I would lean towards taking it out, if only because it adds a bit to the time it takes to run tests, without adding much. But this is only a slight preference.

.gitignore
```diff
@@ -135,3 +135,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# cache file
+url-status-cache
```
mccalluc, line 140: If you need this great, but I'm curious what is different about your environment that it got created at the top level, instead of under `src/ingest_validation_tools` where there's already a git-ignore.

mccalluc (commented):
Thanks for fixing the bad links in the README!

test.sh
lukasz-migas: Indeed - I was using venv in the ivt directory and flake was trying to analyse that.

examples/dataset-examples/good-maldiims/README.md
lukasz-migas: Sure, no problem - I only added it to make sure it wasn't failing

.gitignore
lukasz-migas: It was actually created under `src/ingest_validation_tools` as intended but it was still being tracked. I must admit I didn't see it in the `.gitignore`.

NHPatterson:
This would be very helpful for Vanderbilt TMC to have. We are 95% done with a data submission preparation. For now, I am using this fork to validate our submission in anticipation this will be merged soon.

.gitignore
mccalluc: oh... for whatever reason, you're getting `url-status-cache` while I see `url-status-cache.db`, and the git-ignore is too strict.

For now, just keep this so this branch works for you, but I'll fix it in master, and resolve the redundancy after this is merged.

cebriggs7135:
@NHPatterson Regarding your imminent data submission, the mass spec metadata schema are in revision now per Jeff Spraggins. Please contact me at Christine_briggs@hms.harvard.edu for further information. I want to make sure you use the correct schema. madly-ims will become its with a selection of labeling methods including madly. The schema are not ready yet.

NHPatterson:
Anyway to get this merged? This is a very very minor change to match the file extension convention specified in the imzML specifications in the directory schema. No change to metadata .tsv.

icaoberg:
@NHPatterson I will be reviewing this within the next hour.

icaoberg (approved):
Developer fixed typo in file extension, improved file description and added an example.

icaoberg:
@NHPatterson done.

PR #723, opened 2021-05-13
Mccalluc/status codes

### LINKED ISSUE #142: Clearly distinguish exit status codes
We want to distinguish four ways validate_submission.py can exit:
- status 0: Valid.
- status 1: Unanticipated error
- status 2: Anticipated error (Error message on STDERR)
- status 3: Validation fails (Validation messages on STDOUT)

mccalluc:
Fix #142; @jswelling , I think you're using the python interface, but does this look a reasonable way to be more clear about errors on the command line?

Formerly, validation errors exited with 1: This changes that to 3, because errors with stacktraces from python will return 1.


jswelling (approved):
I do indeed use the python interface and thus will not be impacted by this, but it will be useful to remote users running locally from the command line.

ilan-gold (approved):

PR #726, opened 2021-05-13
more tests

mccalluc:


src/ingest_validation_tools/schema_loader.py
```diff
@@ -212,6 +217,27 @@ def _add_level_1_description(field):
...
 def _validate_level_1_enum(field):
+    '''
+    >>> field = {'name': 'assay_category'}
+    >>> _validate_level_1_enum(field)
+    Traceback (most recent call last):
+    ...
+    KeyError: 'constraints'
+
+    TODO: This should error.
+    Filed https://github.com/hubmapconsortium/ingest-validation-tools/issues/724
```
ilan-gold, line 228: What is this?  There's a TODO that's part of the error?

ilan-gold (approved):

src/ingest_validation_tools/schema_loader.py
mccalluc: That's meant to describe the test below (`field['constraints'] = {}`) ... It should error, but it doesn't.

PR #732, opened 2021-05-18
"submission" to "upload"

### LINKED ISSUE #731: "submission" -> "upload"

mccalluc:
This is mostly just search and replace, so you really don't need to look at every file, but I want to be sure you're both on board with this change.

Fix #731


mccalluc:
Verbal confirmation received at Curation meeting.

PR #738, opened 2021-05-19
factor out number_of_channels

mccalluc:


mccalluc:
Bug in script: will fix and re-run.

PR #753, opened 2021-05-19
factor out bulk_transposition_input_number_nuclei

mccalluc:


mccalluc (commented):
Handled by PR #839

PR #824, opened 2021-05-19
factor out sc_isolation_enrichment

mccalluc:


src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_enrichment.yaml
```diff
@@ -0,0 +1,19 @@
+# scrnaseq-hca-v0.yaml
+- name: sc_isolation_enrichment
+  constraints:
+    required: true
```
cebriggs7135, line 4:
```suggestion
       enum:
           - none
           - FACS
```

cebriggs7135: This should have the same enums as the other sc assays.

cebriggs7135 (changes requested):
Give HCA the same enums for this field.

src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_enrichment.yaml
```diff
@@ -0,0 +1,19 @@
...
+  constraints:
+    enum:
+      - none
+      - FACS
+
+# scrnaseq-v0.yaml
+- name: sc_isolation_enrichment
+  description: The method by which specific cell populations are sorted or enriched.
+  constraints:
+    required: False
```
mccalluc, line 19:
```suggestion
```

src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_enrichment.yaml
```diff
@@ -0,0 +1,19 @@
+# scrnaseq-hca-v0.yaml
+- name: sc_isolation_enrichment
+  constraints:
+    required: true
+  description: The method by which specific cell populations are sorted or enriched.
+
+# scatacseq-v0.yaml; scatacseq-v1.yaml; scrnaseq-v1.yaml; scrnaseq-v2.yaml
```
mccalluc, line 7:
```suggestion
```

PR #827, opened 2021-05-19
factor out sc_isolation_quality_metric

mccalluc:

### DIFF
--- a/src/ingest_validation_tools/table-schemas/assays/scrnaseq-hca-v0.yaml
+++ b/src/ingest_validation_tools/table-schemas/assays/scrnaseq-hca-v0.yaml
@@ -124,12 +124,7 @@ fields:
   constraints:
     required: true
   description: The method by which specific cell populations are sorted or enriched.
-- name: sc_isolation_quality_metric
-  constraints:
-    required: false
-  description: A quality metric by visual inspection prior to cell lysis or defined
-    by known parameters such as wells with several cells or no cells. This can be
-    captured at a high level.
+# include: ../includes/fields/sc_isolation_quality_metric.yaml
 - name: sc_isolation_cell_number
   constraints:
     required: false
--- a/src/ingest_validation_tools/table-schemas/assays/scrnaseq-hca-v0.yaml
+++ b/src/ingest_validation_tools/table-schemas/assays/scrnaseq-hca-v0.yaml
@@ -125,6 +125,8 @@ fields:
     required: true
   description: The method by which specific cell populations are sorted or enriched.
 # include: ../includes/fields/sc_isolation_quality_metric.yaml
+  constraints:
+    required: False
 - name: sc_isolation_cell_number
   constraints:
     required: false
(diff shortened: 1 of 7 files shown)


src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_quality_metric.yaml
```diff
@@ -0,0 +1,15 @@
+# scrnaseq-hca-v0.yaml
+- name: sc_isolation_quality_metric
+  constraints:
+    required: false
```
cebriggs7135, line 4: Since sc_isolation_quality_metric is not required (constraint: FALSE) for HCA but TRUE for other data sources, shouldn't both of these sections to be left as is?

mccalluc:
@cebriggs7135 : scatacseq adds
```
"OK" or "not OK", or with more specificity such as "debris", "clump", "low clump".
```
- Should this be preserved?
- Should there be an enum?

cebriggs7135:
> @cebriggs7135 : scatacseq adds
> 
> ```
> "OK" or "not OK", or with more specificity such as "debris", "clump", "low clump".
> ```
> 
>  * Should this be preserved?
> * Should there be an enum?

Yes, this field should be preserved but no enumerated input is necessary.

src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_quality_metric.yaml
mccalluc: No: The `required: false` will override the base definition.

src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_quality_metric.yaml
```diff
@@ -0,0 +1,15 @@
...
+  description: A quality metric by visual inspection prior to cell lysis or defined
+    by known parameters such as wells with several cells or no cells. This can be
+    captured at a high level.
+
+# scrnaseq-v0.yaml; scrnaseq-v1.yaml; scrnaseq-v2.yaml
+- name: sc_isolation_quality_metric
+  description: A quality metric by visual inspection prior to cell lysis or defined by known parameters such as wells with several cells or no cells. This can be captured at a high level.
+
+# scatacseq-v0.yaml; scatacseq-v1.yaml
```
mccalluc, line 13:
```suggestion
```

src/ingest_validation_tools/table-schemas/assays/scrnaseq-hca-v0.yaml
```diff
@@ -124,12 +124,7 @@ fields:
   constraints:
     required: true
   description: The method by which specific cell populations are sorted or enriched.
-- name: sc_isolation_quality_metric
-  constraints:
-    required: false
-  description: A quality metric by visual inspection prior to cell lysis or defined
-    by known parameters such as wells with several cells or no cells. This can be
-    captured at a high level.
+# include: ../includes/fields/sc_isolation_quality_metric.yaml
```
mccalluc, line 127:
```suggestion
# include: ../includes/fields/sc_isolation_quality_metric.yaml
  constraints:
    required: False
```

mccalluc:
Merging to master: Will regen there and then merge to main.

PR #828, opened 2021-05-19
factor out sc_isolation_tissue_dissociation

mccalluc:


src/ingest_validation_tools/table-schemas/includes/fields/sc_isolation_tissue_dissociation.yaml
```diff
@@ -0,0 +1,9 @@
+# scrnaseq-hca-v0.yaml
+- name: sc_isolation_tissue_dissociation
+  constraints:
+    required: true
+  description: The method by which tissues are dissociated into single cells in suspension.
+
+# scatacseq-v0.yaml; scatacseq-v1.yaml; scrnaseq-v0.yaml; scrnaseq-v1.yaml; scrnaseq-v2.yaml
```
cebriggs7135, line 7:
```suggestion
```

cebriggs7135 (changes requested):

PR #829, opened 2021-05-19
factor out segment_data_format

mccalluc:


mccalluc (commented):
Handled by PR #839

PR #839, opened 2021-05-19
Mccalluc/fields with only one variant

mccalluc:
This replaces a number of PRs I created earlier today where there is only variant, plus this:
```diff
- description: Reagent kit used for sequencing. NovaSeq6000 for example
+ description: Reagent kit used for sequencing
```
Could add it back in, if that would be better?

Judged only by the number of lines, this is making things a lot simpler, and will ensure that tweaks in the future will be applied to all the relevant assay types... The downside is that this is moving us farther away from normal yaml and increases reliance on the inclusion machinery. Though, since the build yaml is checked in under `docs/` it's still easy to see changes for one assay over time.


mccalluc (commented):
Handled by PR #839

ilan-gold (approved):
This makes sense at a high level although it did take me a second to realize what the `# include` was for.  That being said, it's annoying you can't nest yaml, so I don't see another way.

mccalluc:
> a second to realize what the `# include` was for

Changing it to something else would be easy... but I don't think I want to change it to something that would cause lots of syntax errors, and this seems concise. Shrug.

cebriggs7135 (approved):

PR #845, opened 2021-05-20
Mccalluc/loop reference errors reorder

mccalluc:
Reorder the methods, and add some headers for clarity. I know that that `__` has particular semantics in python, but thought it might be useful to distinguish the private methods from the private-private methods? Not sure it's the best idea.

Extends #844... but separate PR to keep that one readable.


ilan-gold (approved):
I'm a fan of the double `_`

PR #848, opened 2021-05-24
forbid NA; code / update fixtures / regen docs

### LINKED ISSUE #840: Disallow "NA", "N/A", etc.
People should not provide a value when there is no value. It would be easy to miss this in the middle of a TSV. This could be another custom constraint that is applied to all fields.

mccalluc:
For a long time we have made clear in the documentation that fields without values should just be left blank. [For example](https://hubmapconsortium.github.io/ingest-validation-tools/scrnaseq-hca/#donor_id):
> HuBMAP Display ID of the donor of the assayed tissue. Leave blank if not applicable.

Since there is still confusion on this, with this PR, there will now be a validation failure if `NA` or any variant is used in any string field.

- @cebriggs7135 : Does this seem right, and will `NA` and `N/A` in bother upper and lower be enough, or are there other variants to check?
- @ilan-gold : Code review

Fix #840


cebriggs7135 (approved):
I think you got the likely permutations of not applicable.

src/ingest_validation_tools/schema_loader.py
```diff
@@ -374,11 +377,16 @@ def _add_constraints(field, optional_fields, offline=None, names=None):
...
+    # In the src schemas, set to False to avoid limit on sequences...
     if field['custom_constraints'].get('sequence_limit', True):
         field['custom_constraints']['sequence_limit'] = 3
     else:
         del field['custom_constraints']['sequence_limit']
+    # ... or to allow "N/A":
+    if field['custom_constraints'].get('forbid_na', True):
+        field['custom_constraints']['forbid_na'] = True
+    else:
+        del field['custom_constraints']['forbid_na']
```
ilan-gold, line 389: Why `del` it?

ilan-gold (approved):

src/ingest_validation_tools/schema_loader.py
mccalluc: I think I just want to minimize the variations that will be seen downstream: Either it's there, and true, or it's not there at all. Could be argued.

PR #850, opened 2021-05-25
Factor out status codes

mccalluc:
- Define the status codes just once
- Give the file a better name: `cli_utils`.


mccalluc:
Erg: namedtuple doesn't have defaults kwarg in 3.6. Will fix.

ilan-gold (approved):

PR #854, opened 2021-05-26
Mccalluc/separate validate tsv

mccalluc:
- Get rid of `validate_sample.py`...
- and replace it with a `validate_tsv.py`, that can validate any of our TSVs in isolation (This is new: In the past, we didn't have a way to just validate an antibodies TSV, though that has been requested.)
- ... so `validate_upload.py` can now just do one thing: Validate entire upload directories.


cebriggs7135 (approved):
This is great.

ilan-gold (approved):

PR #863, opened 2021-06-04
Update Lightsheet z axis specs, # of sections & increments

cebriggs7135:
Seth Curlin revised the Z resolution fields to more accurate terms for the z dimension in lightsheet image capture:
range_z_value & unit - total z distance captured
step_z_value - number of sections captured
increment_z_value & unit - distance between sections


cebriggs7135 (commented):
RE: your comment in Slack, I emailed Seth & Hunter to ask if they want to use this new schema or stick with the original currently in Globus with the 3 light sheet datasets. If so, they will prepare new metadata TSVs, validate & replace the ones in Globus.

PR #864, opened 2021-06-05
Mccalluc/bad network

### LINKED ISSUE #862: Frictionless errors for 'resource temporarily unavailable' don't say which resource
We are seeing errors like the following from ingest-validation-tools table_validator:
>   File "/hive/users/hive/hubmap/hivevm193-prod/ingest-pipeline/src/ingest-pipeline/submodules/ingest-validation-tools/src/ingest_validation_tools/table_validator.py", line 27, in get_table_errors
>    assert len(report['errors']) == 0, f'report has errors: {report}'
> AssertionError: report has errors: {'version': '4.0.0', 'time': 0.103, 'errors': [{'code': 'task-error', 'name': 'Task Error', 'tags': ['#general'], 'note': '[Errno 11] Resource temporarily unavailable', 'message': 'The task has an error: [Errno 11] Resource temporarily unavailable', 'description': 'General task-level error.'}], 'tasks': [], 'stats': {'errors': 1, 'tasks': 0}, 'valid': False}

(The line number has changed in the current code version but the mechanism remains the same).  Some resource being used to validate URLs is not available, but we can't tell from the output which one.

mccalluc:
Fix #862... though this caches the network errors just like successes, and I'm not sure that will be best in the long run. (Remember that `--clear_cache` is available in the CLI.) Once you know which service is causing the problem, I'm not sure what the next step would be.
- Reach out to the third party... perhaps they'd like us to throttle requests, or they can suggest a better way to get the info we need?
- If it's really sporadic, we could add a wait-retry loop on our side.
- If outages tend to last for some period of time, maybe we just can't rely on this service and Chris should remove the check from the schema? 
- To disable network checks at runtime, `--offline` could be used.


jswelling (approved):
Thanks; this does look like what's needed.  I agree that the next steps are still ambiguous.

cebriggs7135 (commented):
Requesting clarification of the proposed changes.

mccalluc:
> Requesting clarification of the proposed changes.

Formerly (#862) if there was a network problem while doing a network check, it would exit with a stack trace. Now, the error message is included in the validation report... So validation still fails, but it's more clear that the problem is with the upstream service, and not a bug in our software.

cebriggs7135 (approved):

PR #868, opened 2021-06-08
update to Lightsheet version 2 schema description.

cebriggs7135:
This version has 5 new fields so the description now describes the changes made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.0.12 - in progress
 - Network problems in report, instead of quitting with stack trace.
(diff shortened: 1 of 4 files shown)


mccalluc:
@khanshawPSC : No
- This PR is just adding a comment that was missed the first time.
- The [PR that made the real change](https://github.com/hubmapconsortium/ingest-validation-tools/pull/863/files#diff-5b82892713208bbdf78686235f022e73b97e25e293c51eb23e75b61bfd808c04) introduced a new version of the schema, so data providers will need to explicitly say version `2` to use the new columns.

PR #871, opened 2021-06-09
remove nanoDESI + deprecation machinery

### LINKED ISSUE #866: Deprecate nanoDESI ... first case of deprecation
- Should still be in generated docs.
- Will not be referenced as a possible schema for validation... which enables the name to be used elsewhere.

mccalluc:
Fix #866 
- Keeping as draft until we can get confirmation that both nanoDESI and nanoPOTS in their current form will be deprecated.
- But separately from that, are the [generated docs](https://github.com/hubmapconsortium/ingest-validation-tools/pull/871/files?short_path=c061adb#diff-c061adb7cbcbb5220066ac8c0940eac17b6fd633decab67c0988b4f5bd00075e) good?
- and is the [validation error message](https://github.com/hubmapconsortium/ingest-validation-tools/pull/871/files?short_path=c061adb#diff-0937f96eeeff57719b50e017093791487c4c1a10559218989885b8623d9efc6fR5) sufficient?


ngehlenborg (approved):

cebriggs7135:
Update: nanoPOTS can be moved to IMS so the nano schema may be deprecated after all.

cebriggs7135 (approved):
Final decision: both nanoDESI & nanoPOTS are moving to the new IMS schema.

PR #890, opened 2021-06-22
Explain  acronymns

mccalluc:
proposed language describing the assay name in English.


src/ingest_validation_tools/table-schemas/assays/af-v1.yaml
```diff
@@ -1,4 +1,5 @@
 doc_url: https://portal.hubmapconsortium.org/docs/assays/af
+description_md: AF stands for autofluorescence microscopy.
```
mccalluc, line 2:
```suggestion
description_md: This schema is for autofluorescence (AF).
```

PR #894, opened 2021-06-22
missing + in slideseq schema

jamestwebber:
this prevented our R1 fastq files from matching.


jamestwebber:
I guess this will need to be fixed in the GH pages, not sure how that works.

mccalluc:
@icaoberg -- Because of time pressure, I'll go ahead and merge this. Hope that's ok.

PR #897, opened 2021-06-28
Use JSON for URL cache

### LINKED ISSUE #896: Delete cache if load fails?
@jswelling reports:
> we're having a problem being reported by frictionless inside table_validator at line 25: ... .  The error frictionless is reporting is 'db type is dbm.gnu, but the module is not available' , for which stackoverflow suggests that there is some sort of caching issue.  If you have any idea about how to deal with it I'd love to hear it.

Maybe just delete the cache file if it can't be read? ... but it would be good to understand what's causing it ...

mccalluc:
Fix #896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.12 - in progress
+- Make the network cache file JSON, for portability.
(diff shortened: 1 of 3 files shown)


mccalluc:
Replaced by #899.

PR #902, opened 2021-06-29
Last changes from master

mccalluc:
Changes suggested by Chris.


mccalluc:
@cebriggs7135 -- This fell through the cracks... Is it still something you want?

PR #909, opened 2021-07-14
Update to description text for lc-ms v2

cebriggs7135:
Description text added on splitting datasets involving multiple analyses.


src/ingest_validation_tools/table-schemas/assays/lcms-v2.yaml
```diff
@@ -1,7 +1,8 @@
 doc_url: https://portal.hubmapconsortium.org/docs/assays/lcms
 description_md: 
   This schema is for liquid chromatography mass spectrometry (LCMS). v2 adds `mass_resolving_power`, `mz_resolving_power`, `ion_mobility`,
-  `spatial_type`, `spatial_sampling_type`, `spatial_target`, and `resolution_{x/y}_{value/unit}`.
+  `spatial_type`, `spatial_sampling_type`, `spatial_target`, and `resolution_{x/y}_{value/unit}`. In the case of datasets in which more
+  than one analyte_type was interrogated (eg. lipids plus metabolytes), those datasets should be split into one dataset per analyte.
```
mccalluc, line 5:
```suggestion
  than one `analyte_type` was interrogated (e.g. lipids plus metabolytes), those datasets should be split into one dataset per analyte.
```

PR #912, opened 2021-07-19
Mccalluc/error unexpected dirschema flags

### LINKED ISSUE #910: Scan for unexpected keys in directory schema
A `ignore` flag was used by mistake: we should make sure only recognized flags are used.

mccalluc:
Fix #910. Directory schemas with unexpected keys will fail at doc-generation time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.12 - in progress
+- Catch unrecognized keys in dir schema.
 - Ammend LCMS docs.
 - Fix CI: point to right branch.
 - Document Donor and Sample Metadata process.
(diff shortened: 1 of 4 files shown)


icaoberg (approved):
Looks fair.

However, @mccalluc what is the purpose of the `example` key? What it will be used for?

From what I understand, the valid keys are now

```
{'pattern', 'required', 'description', 'example', 'is_qa_qc'},
```

mccalluc:
@icaoberg :
> what is the purpose of the `example` key? What it will be used for?

Looks like it's not being used much, but there is an example [here](https://hubmapconsortium.github.io/ingest-validation-tools/codex/): The idea is that if you have a complicated pattern, an example can be supplied, since that might be easier to read. The example will be validated against the pattern, to make sure it actually works.

PR #915, opened 2021-07-22
Update index.md

pecan88:
Change request to:
- add current pipeline version 
- changed title to move SNARE to end and focus on ATAC element

Please note: we'd want to change [the same text string located on this page](https://hubmapconsortium.github.io/ingest-validation-tools/) under the "Sequencing assays" category.  From looking at the index, it appears to auto generate that list, so I'm refraining from touching any code ...

### DIFF
--- a/docs/scatacseq/index.md
+++ b/docs/scatacseq/index.md
@@ -1,5 +1,5 @@
 ---
-title: SNARE-seq2 / scATACseq / sciATACseq / snATACseq
+title: scATACseq / sciATACseq / snATACseq / SNARE-seq2
 schema_name: scatacseq
 category: Sequence assays
 all_versions_deprecated: False
@@ -883,3 +883,9 @@ Relative path to file or directory with instrument data. Downstream processing w
 | required | `True` |
 
 </details>
+
+## Pipeline
+
+<details markdown="1" open="true"><summary><b>Version 1.4.3 (current)</b></summary>
+
+[Pipeline package](https://github.com/hubmapconsortium/sc-atac-seq-pipeline/releases/tag/v1.4.3)


mccalluc:
@pecan88 :

> changed title to move SNARE to end and focus on ATAC element

I think I can change the order of the list this comes from, but I wouldn't encourage us to pin to much on the order of these.

> add current pipeline version

Can you give me the story that having this information here satisfies? On the call this morning, we talked a little bit, and it sounded like you might believe that the TMCs should be running the pipeline locally to validate before submission. There was a consensus that we are _not_ asking the TMCs to do that... but maybe we do need checks on the files that go beyond what the directory schema covers? I believe Kevin said he'd talk with you, so wait for that before replying.

mccalluc:
<details><summary>From Peter</summary>

> Developers, PIs, and consortium stakeholders can all give the same answer to the question "under which version are we publishing this assay-specific dataset?"  (Likely we'd display the versions used within the portal as well, but that's possibly getting ahead of ourselves.)
>
>Operationally, the three required element are metadata schema (already versioned & well communicated by Chuck), directory structure (need to add version, need to improve communication) , & pipeline (have versions in GitHub, now need to link with above two elements and communicate more clearly that these are the requirements / deliverables needed for a new assay).
>
>So, ideally we have the versions for each assays' above three elements including "current version" identified and ideally the display presented to the audiences above is "as closely coupled to the technical implementation / code as possible".  (i.e. Reduce number of points of failure for communication & understanding.)

</details>

mccalluc:
Replaced by #919

pecan88:
@mccalluc - thanks re: the order change.  Rationale is that we talk about RNA-seq and ATAC-seq very frequently in the context of data publication & reporting.  SNARE contains both.

Re: story that having pipeline version information displayed will satisfy:
1. As a data provider, I want to know which pipeline is most current / appropriate for my data type so that I can ensure the processing produces high quality results that satisfy me
2. As a project manager, I want to know which pipeline is most current and validated to work for a given data type so that data providers have a stable, working instance to test against and so that we reduce the number of cycles required to get datasets out the door
3. As a HuBMAP senior leader, I want to know the current version (relative to previous ones) so that I can brag about how much progress my consortium makes compared to other consortia and so that I can see problem areas that affect our data production
4. As a developer, I want a clear summary of the required technical products to publish data so that I understand how the process works and so that I can onboard new developers quickly and correctly

While I'm sure there's plenty of detail and clarifications we could call out based on the stories above, I'm hoping this will communicate the rationale supporting the relevance of showing the pipeline version.

PR #919, opened 2021-07-22
Mccalluc/pipeline info

mccalluc:
Replaces Peter's #915 -- I've moved it up to the directory schema since it's about the structure of the files, and give a little explanation. 

- @pecan88: confirm that the [input structure](https://github.com/hubmapconsortium/ingest-validation-tools/pull/919/files#diff-6873595a53fb320c8325fd60939048a58e29987172d8788965beace1fdca493a) is sufficient (there's a lot of other info in your spreadsheet, but if it doesn't need to be captured here, great), and the [output presentation](https://github.com/hubmapconsortium/ingest-validation-tools/pull/919/files#diff-6529bbf4820ed6a095cfbc38743ee532786b64be8bec652f17bb0dc570c68dea) is good.
- @ilan-gold: code review.
- @jswelling or @mruffalo: Confirm that you agree this information should be presented to data providers here, and that you will be responsible for updating the yaml and regenerating the docs, to keep it in sync the the actually-deployed pipeline versions; If neither of you can agree to that, let's find time to talk.


pecan88 (approved):
@mccalluc - so based on my non-expert review, these changes look good.

One feedback item for your consideration as to whether it's worth it or not:
When linking to the repo with the pipeline present, would it be preferable to go to the tags page that lists versions?  Or safe to assume folks will figure out on their own?

pecan88 (approved):
@mccalluc  - based on my non-expert review, I think this looks good.

Only potential consideration is whether link to the repo's parent page or to the page listing the versions.  I'll defer to you on what's best given your expertise.

mccalluc:
@jswelling : More info from PK about the motivation, just to have it in one place:

> here's where I'm trying to get us:
>
> Developers, PIs, and consortium stakeholders can all give the same answer to the question "under which version are we publishing this assay-specific dataset?"  (Likely we'd display the versions used within the portal as well, but that's possibly getting ahead of ourselves.)
>
> Operationally, the three required element are metadata schema (already versioned & well communicated by Chuck), directory structure (need to add version, need to improve communication) , & pipeline (have versions in GitHub, now need to link with above two elements and communicate more clearly that these are the requirements / deliverables needed for a new assay).
>
> So, ideally we have the versions for each assays' above three elements including "current version" identified and ideally the display presented to the audiences above is "as closely coupled to the technical implementation / code as possible".  (i.e. Reduce number of points of failure for communication & understanding.)
>
> The reason I am thinking of the page you created is because it does that close alignment better than anything else I've seen in HuBMAP.

mruffalo:
@mccalluc I'm on board with keeping the yaml files up-to-date with new pipeline releases, but I don't know what "regenerate the docs" means. I've skimmed the implementation, and this isn't clear to me: does this assume that every data type is processed by exactly one pipeline? That assumption isn't valid; we split CODEX data processing into two pipelines: Cytokit + our data collection scripts, and SPRM -- and these are versioned independently.

mccalluc:
@mruffalo :

> I don't know what "regenerate the docs" means

Though the docs are on [GH Pages](https://hubmapconsortium.github.io/ingest-validation-tools/), the processing goes beyond what's feasible with Jekyll (like making Excel data entry templates), so generated documentation is checked in to this repo in the `docs/` directory. Usually running `src/generate_docs.py your-assay docs/your-assay` will suffice: CI will error if anything else is needed, or if there is a problem with the YAML.

> we split CODEX data processing into two pipelines

Thanks for pointing that out. I'll generalize the data structure to be a list, instead of single object.

jswelling:
I want to better understand the purpose of this- the minor version numbers of the workflows change very frequently, and many of those changes have no impact on existing datasets.  I'm also unclear on why ingest-validation-tools is the right place for this. Is the argument that a lot of detailed documentation needed by the data providers is already there, so we might as well put this info in the same place?
In practice, when the data provider completes the data for submission, the workflow that will be used to generate secondary datasets has not yet been defined- and we expect to run the same submitted data through different versions of the workflows going forward, without any contact back to the data provider beyond normal QA.

mccalluc:
@pecan88 : ☝️ Get in touch with Joel. I'm going to remain agnostic about whether this is the best way to communicate the information, but will try to support whatever the consensus decision is.

src/ingest_validation_tools/docs_utils.py
```diff
@@ -469,4 +487,24 @@ def _make_dir_description(dir_schema):
         row.append(description_md)
 
         output.append('| ' + ' | '.join(row) + ' |')
-    return '\n'.join(output)
+    table = '\n'.join(output)
+    pipeline_infos_md = ' and '.join(make_pipeline_link(info) for info in pipeline_infos)
+    pipeline_blurb = \
```
ilan-gold, line 492: So we want this written out with or without a pipeline part of the docs for an assay?

ilan-gold: It seems we only have one of these pipeline info blurbs

mccalluc: > So we want this written out with or without a pipeline part of the docs for an assay?

No: The `pipeline_blurb` will be be empty if there are no pipelines. See below: `if pipeline_infos else ''`

> It seems we only have one of these pipeline info blurbs

Yes, for now: Didn't want to invest too much here when it's not clear that this has buy-in.

mccalluc:
@ilan-gold 
> If the pipeline info is mandatory, perhaps it makes sense to add info now for all assays?

It's not mandatory... None of the other generated docs change with this PR, so I'm confident it doesn't change the generated doc unless it is included.

pecan88:
@mccalluc - I spoke with Joel & IEC folks and @jswelling and @mruffalo met this afternoon to discuss the versioning.  At this point in time, my understanding is that the approach is to declare a current working version for 1) metadata specification (you've already covered ), 2) directory structure (Ivan will need to indicate version number), and 3) pipeline version that a data provider should use for testing to see if their data is processed successfully.  I anticipate infrequent changes to metadata & directory structure specs and potentially frequent changes to the current pipeline version.

The rationale behind having this information on the pages you put together are because they are so good and having all of the requirements in one place will help data providers, program officers outside the consortium interest in how we run things, and HuBMAP stakeholders to understand "how the data sausage is made".  :-)

mccalluc:
@pecan88 

> pipeline version that a data provider should use for testing to see if their data is processed successfully

When this came up at stand up, the consensus was that we were **not** expecting data providers to be running locally.
- Has the consensus changed?
- Or, are you working to change the consensus?
- Or, do you mean "could" instead of "should"?

> directory structure (Ivan will need to indicate version number)

Version numbers on directories is a new thing, and I'm not sure it's necessary. Filed #923.

jswelling:
In conversation with Jonathan, he favors having data providers run test workflows locally (in addition to running upload validation).  I do not favor requiring the data providers to run workflows.

mccalluc:
@jswelling prefers that all the pipelines be in a single file, for ease of maintenance.

mccalluc:
@jswelling -- pipeline info in a single file. Please approve, if you approve.

mruffalo (approved):
Looks good to me.

jswelling (approved):

PR #921, opened 2021-07-23
Mccalluc/tweak enums

mccalluc:
@ilan-gold -- This started as a simple addition  to enums, but then when I tried to regenerate, the files for `sample` weren't updated. Digging into it:
- For assays, the TSVs are named like `af-metadata.tsv` but for non-assays, including `sample`, the name is just `sample.tsv`
- The loop in the bash script was skipping non-assays: `[ -e docs/$D/$D-metadata.tsv ] || continue`
- Considered making the names more consistent... but I think that would cause more confusion. It's be useful to distinguish between the "metadata TSVs" and other TSVs, like `antibodies.tsv` or `contributors.tsv`.
- After I fixed the bash script, all of the non-assays were updated, at to change the urls from `master` to `main`.

So I think this is good now, but I want you to understand the landscape.


tests/test-generate-docs.sh
```diff
@@ -24,9 +24,10 @@ rm -rf docs-test
 for TYPE in $(ls -d docs/*); do
   # Skip directories that are unpopulated:
   TYPE=`basename $TYPE`
-  LOOKFOR="docs/$TYPE/$TYPE-metadata.tsv"
-  if [ ! -e $LOOKFOR ]; then
-    echo "Skipping $TYPE. To add: touch $LOOKFOR"
+  LOOKFOR_ASSAY="docs/$TYPE/$TYPE-metadata.tsv"
+  LOOKFOR_OTHER="docs/$TYPE/$TYPE.tsv"
```
ilan-gold, line 28: Is there something to call this other than `OTHER` - are these entities?

ilan-gold (approved):

tests/test-generate-docs.sh
mccalluc: There probably could be a better name, but "other" is at least consistent: `get_other_schema`, `table-schemas/others`, and on the GH Pages homepage, `Other TSVs`. These are the TSVs that aren't _assay metadata_ TSVs: ie, contributors, antibodies, samples.

ilan-gold: Sounds good

PR #925, opened 2021-07-28
Added description_md for CEMS

cebriggs7135:
"description_md: This schema is for capillary electrophoresis - mass spectrometry (CEMS)."


docs/cems/v1.yaml
```diff
@@ -1,5 +1,6 @@
 # Generated YAML: PRs should not start here!
 doc_url: https://portal.hubmapconsortium.org/docs/assays/maldi-ims
+description_md: This schema is for capillary electrophoresis - mass spectrometry (CEMS).
```
mccalluc, line 3: @cebriggs7135 : See note on line one. PRs should never begin in `docs/`.

mccalluc (changes requested):

PR #926, opened 2021-07-28
Updated CellDive directory structure schema

icaoberg:
Update to CellDive directory structure schema. Tested on a copy of one dataset

```
bash ./ge-celldive1.sh

No errors!
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.13 - in progress
+- Updated CellDive directory structure.
 - Updated CODEX directory structure.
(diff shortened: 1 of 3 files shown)


jswelling (changes requested):
It looks like some of the regex fields you are encoding are integers.  If so, you should probably use the \d regex field type.  For example, you can encode something like 1234_56.png as  '\d+_\d+\.png' .  The \d+ parts mean 'one or more decimal digits'.  I suggest this change on the theory that it's always good to be specific if we can.

jswelling (approved):

jswelling (approved):

PR #928, opened 2021-07-29
Mccalluc/directory schema versions

mccalluc:
Directory schema versioning. 
- Data directories may now include `directory-schema-version.txt`: If present, it will be read, and the specified version will be used for validation.
- If missing, assume `v0`. (Analogous to metadata schema versioning.)
- Directory schemas have been renamed, adding a `-v0` to most of them.
- Internally, directory schemas have been rearranged: They were just a list, and now that list is under `files`.
- This was done so I could add an optional `deprecated` flag.
- I've resurrected the previous version of the CODEX directory schema as `codex-v0.yaml`, and marked it as deprecated, so you can what the docs look like, and I've added a test to show that it fails if a v0 codex comes in.
- I've also added a test of an upload specifying a version that doesn't exist.
- Lots of method renaming, to distinguish metadata and directory schemas.
- I've updated the submission structure diagram, but in the same file, instead of making a copy: So that's up on the README now. Sorry.
- Versions do not need to be numbers, so @icaoberg can simplify the CODEX schema by splitting it into two separate schemas for Stanford and Akoya, so we'll have `directory-schemas/codex-v2-stanford.yaml` and `directory-schemas/codex-v2-akoya.yaml`... or something like that. 

This is more complex, but it's what was requested. I hope I've explained it well enough that I will not be spending time re-explaining it in the future.

- **question**: Is there code that uses the regexes here to provide file descriptions? Or is that only with the pipeline outputs, and the regexes here are not read by any other code?

cc: @ngehlenborg , @cebriggs7135 , @pecan88


jswelling (changes requested):
I'm not sure I understand having the .xlsx files in the docs.  Are these intended as examples?  If so, shouldn't they be .tsv files?

mccalluc:
@jswelling : The excel data entry templates have been part of this project since #440: They have nothing to do with this PR. If you download one, you'll see that it provides:
- a description for each of the columns
- pull downs to select enumerated values
- and range and type checks.

src/ingest_validation_tools/docs_utils.py
```diff
@@ -377,30 +377,54 @@ def _make_toc(md):
     return f'<blockquote markdown="1">\n\n{joined_mds}\n\n</blockquote>\n'
 
 
-def _make_dir_description(dir_schema):
+def _make_dir_descriptions(dir_schemas):
+    sorted_items = sorted(dir_schemas.items(), key=lambda item: item[0])
```
mruffalo, line 381: Minor nitpick: might be worth implementing as
```
sorted_items = sorted(dir_schemas.items(), key=operator.itemgetter(0), reverse=True)
```

(or `key=itemgetter(0)` after `from operator import itemgetter` at the beginning of the file)

mruffalo (approved):
Looks good to me :+1:

mccalluc:
Coming out of meeting August 10:
- Original direction was to make directory schema versions "like" metadata schema versions. That by itself doesn't seem to be a priority any longer...
- Preference is for the directory schema not to be explicit in the dataset: Instead all schemas will be scanned, and if any match, great!
  - ... but the downside is that if none match, we can't say which one _should_ have matched, but Ivan and Joel are ok with this trade-off.
  - For now, the error messages will become more ambiguous.
  - Down the road, try to guess which one should have matched by returning the one with fewer errors.
  - And/or, add a CLI option to explicitly indicate the intended schema
- Add support for macro functionality, like with the metadata, to factor out sets of files that are shared between schemas. This makes it isomorphic to [Joel's proposal](https://docs.google.com/document/d/1ZQgweiIgnVev049Z1ba3PQEf_-lx2RfwTQB2zxrqpaQ/edit).

There are other problems we have with validation, but they don't relate to this PR:
- TMCs have trouble reading regexes, but:
  - regexes will be much simpler when we don't need to encode variations in a single file,
  - Using regex is consistent with what Matt does with the output from pipelines.
  - Switching to glob would loose expressive power... without gaining much in readability.
- TMCs aren't actually saying what files they'll be providing till late in the game, and then they can't stick to their plans. Nothing a validation tool can do to fix this.
- Some assays (nanoDESI) don't seem to be in the right bucket... but at least being able to version schemas will make it easier to handle variation.

khanshawPSC:
@mccalluc it would be good to understand what changes will need to be made so that the data we have that has not been ingested yet can validate properly. After our discussion, I realized these changes could cause the data submitted by the data providers but not ingested yet to fail validation. Will that be the case?

mccalluc:
@khanshawPSC 
> After our discussion, I realized these changes could cause the data submitted by the data providers but not ingested yet to fail validation. Will that be the case?

No, not with the agreed-upon plan, and not in the current state either. Data that's already been submitted will still validate, and it will be up to Ivan to decide if/when a schema is no longer good for validation.

khanshawPSC:
Thanks for that clarification. My concern is no longer.

mccalluc:
@icaoberg / @jswelling : With the latest commits, it now tries to validate with every directory schema for that type, as requested. If there's one that passes, great; otherwise, it tries to pick the the one that was the least-wrong (for which I'm counting the number of lines in the error) and presents that.

cc @ngehlenborg , since you were asking about this on Slack.
cc @pecan88 , since you kicked this off.

My opinion is that this is a mistake, and we will regret the increased code complexity when we try to maintain this, and the users will be confused by the non-obvious relationship between the their input and the error they get. I think either the earlier state on this branch, or the status quo is preferable, but I'd like to step back from this decision. If further tweaks are needed, I think it would be good for someone else to do it, to make I'm not the only one familiar with this codebase.

pecan88:
@mccalluc - I'd suggest we hold activity here until we get additional clarity from key HIVE PIs and stakeholders on the purpose of directory structures, cost/benefit relating to complexity, and ultimate presentation to end users (feedback shared from NIH).  I am hopeful this will facilitate a better solution (ideally with less effort involved) and avoid further (expensive) churn.  Thank you for driving the specificity and critical consideration around implementation and the impacts.

mccalluc:
On September 3, Jonathan said that he would get the TMCs to conform to a single schema. If that is true, versioning is not necessary, or useful. If I'm misunderstanding the outcome of the meeting, I'd like a PI to indicate the way forward.

mccalluc:
(I'd like to keep the branch around, for now.)

PR #932, opened 2021-08-04
Mccalluc/multivalued as comma delimit

### LINKED ISSUE #929: `cell_barcode_read` -> `cell_barcode_reads` or something like that.
Prepare two PRs, one with a JSON list, one comma delimited. We will see more multi-valued fields, so figure out the right way to do this.

mccalluc:
Another possible way to fix #929: Rename field, and value should be comma-delimited.

Plus:
- Simple
- Consistent with other fields.
- Easy to require that there be at least one file listed.

Minus:
- It's far from what they were supplying, and unless manual curation is actually happening, old form data would still slip through.
- It's possible for filenames to contain commas.
- Machine parsing is trickier: Padding space should probably be removed?


cebriggs7135 (approved):
OK

PR #933, opened 2021-08-05
Mccalluc/new sample organ schema

mccalluc:


pecan88:
Here's some attempts at answers to @mccalluc 's questions:

Q: ID field names: I've shortened, for consistency. Ok?

A:  I believe @computationdoc advocated for the longer names. I'll seek to get this on his radar for response.

Q: What's up with protocols? Do we keep collecting them in the UI, or are these moved to the TSV?

A: One thought from our U.Penn SME: "Trying to capture them in the TSV would be very error prone unless the ingest validator can directly validate a DOI from a TSV against protocols.io."

Q: The Excel sheet gives an enumeration: RIN | Histopathology. What precisely should be done with that?

A: The “sample quality criteria” field exists in v0 of the sample TSV. We aren’t advocating handling this field any differently than however it’s currently being handled. Penn is currently planning to provide "histopathology report” as just unformatted text. I’m sorry I don’t have any examples yet. [U.Penn SME]

Q: How/where should organ left vs. right be recorded?

A: I think a new field in the organ schema would be appropriate. I’m sorry the Sample Team missed this when designing the v1 schema. [U.Penn SME]

Q: Many field names correspond to fields in the previous sample schema. Can we just use some of the old names? *_ischemic_* vs *_ischemia_* for example. Or should the old named be mapped to the new ones for consistency at index time? Or does this not matter? 
 
A: These only fields we intentionally renamed are below. [U.Penn SME]
“Sample preparation temperature” was previously "Specimen presevation temperature"
"Sample quality criteria" was previously "Specimen quality criteria"
"Sample pathology distance value" was previously "specimen tumor distance value"
"Sample pathology distance unit" was previously "Specimen tumor distance unit”

mccalluc:
@pecan88 -- I would really, really prefer to have a single consensus response, and not to have a thread here. I'm not sure how to state this more clearly. To correct one misunderstanding:

> Trying to capture them in the TSV would be very error prone unless the ingest validator can directly validate a DOI from a TSV against protocols.io

The software already does for protocols.io DOIs, as well as the IDs particular to antibodies TSVs.

pecan88:
@mccalluc  - the IEC PIs have identified @ropelews as being the right person to review and provide initial answers to these questions.  I'm looping in @khanshawPSC for awareness.

ropelews:
Chuck: I'm preparing a document to share internally with the IEC for a unified response.

mccalluc:
I believe the parse is
> close this PR and (start a new one once the issues have been resolved to everyone's satisfaction)

rather than
> (close this PR and start a new one) once the issues have been resolved to everyone's satisfaction

So I'll close. Apologies if mistaken.

PR #934, opened 2021-08-09
Update the flowchart to reflect the roles of Bill and PK

mccalluc:
Some discussion this morning about why/how particular datasets had gotten published made me realize that the end of the flowchart hasn't been accurate. This adds one additional step to the diagram:
> **PM** gathers approvals;
> **PSC** publishes

It also 
- adds these two roles to the key in the lower right.
- cleans up the arrows at the bottom of the left column.

My understanding is that we're still waiting for the discussion about pinning down who _precisely_ needs to approve before publication, so more tweaks might come out of that, but I think the two questions can be considered separately. (Matt and Ilan have had an ad-hoc role in approval, but I don't believe they are a permanent part of the approval process for each dataset going forward.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.13 - in progress
+- Update the flowchart to reflect the roles of Bill and PK.
 - Hotfix on CellDive directory to reflect changes to dataset.
 - Updated CellDive directory structure.
 - Updated CODEX directory structure.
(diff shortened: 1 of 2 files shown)


ngehlenborg (approved):

ngehlenborg:
@pecan88 @shirey: Here's the  link to the latest version of the flow chart: https://docs.google.com/drawings/d/1fDhORYm8DYnCnbvpIrMMN0OxFQakHir_Ss071q2ySNc/edit

pecan88 (approved):

PR #946, opened 2021-09-10
Mccalluc/tighter booleans

### LINKED ISSUE #945: Booleans should only allow TRUE / FALSE

mccalluc:
- Fix #945.  
- In the process, realized that email format validation was not being applied, and fixed it.

For review, look at the first couple examples under `docs/` and then scroll to the bottom: They are all the same.


mccalluc:
Decided against this.

PR #953, opened 2021-09-27
Add MALDI-IMS-neg and MALDI-IMS-pos to ims table schemata

jswelling:
The new Vanderbilt MALDI-IMS datasets are currently not runnable because the official assay type names are MALDI-IMS-pos and MALDI-IMS-neg while the assay_type constraints in ingest-validation-tools/src/ingest_validation_tools/table-schemas/assays/ims-v2.yaml (and earlier versions) is just MALDI-IMS .  This PR resolves this conflict by adding MALDI-IMS-pos and MALDI-IMS-neg to the table-schemas.

### DIFF
--- a/src/ingest_validation_tools/table-schemas/assays/ims-v0.yaml
+++ b/src/ingest_validation_tools/table-schemas/assays/ims-v0.yaml
@@ -9,6 +9,8 @@ fields:
   constraints:
     enum:
       - MALDI-IMS
+      - MALDI-IMS-pos
+      - MALDI-IMS-neg
 - name: analyte_class
   constraints:
     enum:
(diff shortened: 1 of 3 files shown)


cebriggs7135 (approved):

mccalluc (commented):
Data could still come in, and be validated, that had the old name... and my sense is that the old name doesn't work for you. Is the plan for now to make it possible for people to do the right thing, and latter a new version can be minted (and the current versions deprecated) so people are prevented from doing the wrong thing?

I would like to loop @ngehlenborg in because we haven't been on the same page about what kind of changes require a new schema version... If the overall plan is good, I think this particular change is fine, because it loosens restrictions.

ngehlenborg:
I would like to get clarification on this with the TMCs before proceeding. I will reach out to the parties that can shed light on this and added myself as a reviewer to ensure that this gets sorted out before we move on.

mccalluc:
@cebriggs7135 -- Does your recent approval mean that @ngehlenborg 's concerns been resolved?

mccalluc:
(Converting to draft to avoid accidental merge until @jswelling can confirm that he's on the same page.)

PR #955, opened 2021-09-28
Fix a problem apparently due to type hinting mismatch in importlib

jswelling:


mccalluc (approved):
I'll go ahead and merge... which isn't the usual, but I think we want to move this along.

PR #958, opened 2021-10-01
Upgrade LC-MS directory structure schema

icaoberg:
Added directory structure schema and documentation.


jswelling (approved):
Nothing tricky here, I think

PR #960, opened 2021-10-01
level 1 metadata update

cebriggs7135:
Changed "3" to "4" assay_categories and added imaging_mass_spectrometry to the list.


mccalluc (changes requested):
regenerate docs, and add a changelog.

cebriggs7135 (commented):
Does this mean you made the same edits to the schema_loaded.py file?

mccalluc:
@cebriggs7135 -- Sorry for the ambiguity: I meant that you need to regenerate docs, and add a changelog entry in this branch. Then tests will pass, and this can be merged.

mccalluc (approved):

PR #963, opened 2021-10-06
Mccalluc/check assay terms

### LINKED ISSUE #216: Do assay type lists need to be kept in sync?
There are different assay type lists floating around, like [this one](https://github.com/hubmapconsortium/search-schema/blob/master/defininitions/enums/assay_types.tsv) from Bill, and [this one](https://github.com/hubmapconsortium/ingest-validation-tools/blob/master/src/ingest_validation_tools/table-schemas/level-1.yaml#L43-L69) from Chris.

@shirey : I think the question is for you: Should I/we do anything to keep these lists in sync? If pipeline code never looks in the metadata.tsv then perhaps the exact values here are inconsequential for you? Do you imagine that may change? Conversely, Joel has thought that parentheses would be bound to cause problems (#194)... but if it's just a text field in ES, the emphasis should be on making these readable / idiomatic human-readable strings.

mccalluc:
Draft: Fails now, but when there's an API that gives us a list of the names supported downstream, we can use that. My belief is that this is the only work needed in this repo for this project.

Fix #216.


mccalluc:
@jswelling :
> We don't have support on the search-api side for assay_category or analyte_class. ... should we try to pursue this now?

No: I am not proposing that we do. The CI doctest only covers `assay_type`:
```
>>> local_names = set(shared_enums['assay_type'])
```

>  If so, we should be talking about the code for that CI test, shouldn't we?

This PR _does_ propose a CI test:
```
>>> downstream_names = set(requests.get(
...    'https://search.api.hubmapconsortium.org/assaytype?simple=true&primary=true'
... ).json()['result'])
>>> local_names - downstream_names
```

Maybe it imports typeclient instead of hitting the API -- I really don't care -- However you care to provide that information. Does the intention of this PR make sense now?

jswelling:
Ah, my mistake- I was still thinking in terms of separate test files.  My proposed change would be:
```
>>> local_names = shared_enums['assay_type']
>>> remote_match = []
>>> remote_mismatch = []
>>> for name in local_names:
...     rslt = requests.post('https://search.api.hubmapconsortium.org/assayname', json={'name':name}, headers={'Content-Type': 'application/json'}).json()
...     if 'error' in rslt:
...         remote_mismatch.append(name)
...     else:
...         remote_match.append(name)
>>> remote_mismatch
[]
```

At the moment there are 33 mismatches.  The great majority can be fixed by adding 'alt-names' to the corresponding assays, for example adding '3D Imaging Mass Cytometry' to IMC3D.  A number look like new assay types or unknown variants on older assay types, though, and they will need specific PRs to assay_types.yaml .

jswelling:
I created a matching PR against assay_types.yaml in search-api - see https://github.com/hubmapconsortium/search-api/pull/400

ngehlenborg (approved):
Approved to unblock processing of new assay types. Will still require further discussion to resolve the general problem of remapping of assay types (per discussion during portal call on 11/12/2021).

PR #966, opened 2021-10-18
example directories

### LINKED ISSUE #952: Template directory structure
tool to create empty directory structure for labs to then populate

mccalluc:
Fix #952. I don't care strongly about the basic issue one way or another, but if it's going to be done, this is the way it should be done: We should build from the examples that are already available for some assays, instead of creating a new, inconsistent system. (But I'm ok with this not going ahead at all, too.)

```
$ src/generate_example_directory.py --dir_schema src/ingest_validation_tools/directory-schemas/codex.yaml --target /tmp/foo/bar
Created /tmp/foo/bar/NAV.tif
Created /tmp/foo/bar/summary.pdf
No example for (raw|src_[^/]*)/exposure_times\.txt
Created /tmp/foo/bar/raw/reg_00.png
No example for (raw|src_[^/]*)/[Ee]xperiment\.json
No example for processed/HandE\.tif
No example for processed/HandE_RGB\.tif
No example for processed/HandE_RGB_thumbnail.jpg
No example for (src_[^/]*|drv_[^/]*|extras)/[sS]egmentation\.json
No example for (raw|processed)/config\.txt
No example for (raw|processed)/config\.txt|(src_[^/]*|drv_[^/]*|extras)/[sS]egmentation\.json
No example for src_[^/]+/channelnames_report\.csv
No example for (raw|src_.*)/[cC]yc.*_reg.*/.*_Z.*_CH.*\.tif
No example for src_.*/cyc.*_reg.*_.*/.*\.gci
No example for (raw|src_.*)/channel[Nn]ames\.txt
No example for (raw|src_.*)/.*
No example for (processed|drv_[^/]*)/.*

$ tree /tmp/foo/bar
/tmp/foo/bar
├── NAV.tif
├── raw
│   └── reg_00.png
└── summary.pdf
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.13 - in progress
+- Script to populate directory based on examples.
 - Updated LC-MS directory structure schema.
 - Work around mypy importlib type hinting problem.
 - Longer assay description for LCMS, and supporting machinery.
--- a/script-docs/README-generate_example_directory.py.md
+++ b/script-docs/README-generate_example_directory.py.md
@@ -1,0 +1,10 @@
+```text
+usage: generate_example_directory.py [-h] --dir_schema DIR_SCHEMA --target
+                                     TARGET
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --dir_schema DIR_SCHEMA
+                        Directory schema file
+  --target TARGET       Target directory to populate
+```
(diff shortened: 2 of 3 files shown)


ngehlenborg:
@mccalluc Can you clarify what it means to build from examples? I don't know where the examples are coming from, how complete they are, and how they are maintained (if at all).

If I understand the terminal output above correctly, only a subset of the "expected" directory structure was created, presumably because only a few of the files were available in the example. Is that correct?

mccalluc:
@jswelling / @cebriggs7135 / @icaoberg -- My understanding is that there is no longer a need for example dataset directories. Closing.

PR #970, opened 2021-10-21
Add label_name field to labeling "includes"

cebriggs7135:
We have a "labeling" field that takes a boolean but if the user enters TRUE, we then need a place to describe the label. TMT is an LC-MS assay type that uses tandem mass tags to label individuals samples so they can be multiplexed. We're about to accept a large TMT collection off TMT datasets from Stanford.

However, if this change in an "includes" file will affect other metadata schema, should this be added to the LC-MS separately so the other docs don't have to be regenerated? I think it's the only assay with a labeling field, though.


mccalluc:
- I believe Nils has stated that a change to the set of columns requires a new version. Is this an exception? Is there a plan to deprecate the previous version?
- Updates to schemas require regeneration of the documentation.

PR #978, opened 2021-10-25
Mccalluc/mibi

### LINKED ISSUE #967: Implement MIBI schema
Based on
https://docs.google.com/spreadsheets/d/1svITiu82IC556PQmBy9kuizXbVDjZlPZFhJe2xet2qs/edit#gid=1468496505

@cebriggs7135 :
- Which of these fields are required?
- What are the allowed enumerations?
- Please make the types explicit.
- Please provide the descriptions as a separate column.

mccalluc:
Fix #967. The rendered documentation can be seen [here](https://github.com/hubmapconsortium/ingest-validation-tools/pull/978/files?short_path=a620559#diff-a620559a741d7b4e8ee54d811476bae33d31ee1c2a322a3de0304cd5c7103214).

- For a number of fields, the required flag is different from what other other schemas used, starting with `resolution_x_value` which is optional in the spreadsheet, but required on other assays. For now, this PR uses the same setting as for other assays.
- Is a documentation url available?
- Should we wait for the directory schema, since we've agreed that the metadata and directory schemas will be developed in tandem?
- Why does this schema not have an acquisition instrument?

On the original issue I noted a number of problems in the spec, including:

- The enum for `max_x_width_unit` is given as `pixels`: I believe this is an error. In other places where this field is used, a real-world distance unit is spcified: `um`.
- `number_of_antibodies` is given twice.
- `Primary ion current unit, typically nA or pA`, but the only enum given is `nA`.
- The allowed values for `assay_category` have underscores.
- I will capitalize `multiplex ion beam imaging` for consistency.


cebriggs7135:
The final schema is here:
https://docs.google.com/spreadsheets/d/1svITiu82IC556PQmBy9kuizXbVDjZlPZFhJe2xet2qs/edit#gid=1468496505

I just added one enum to area_normalized_ion_dose_unit: the enums are now nA*hr/mm2 , pA*hr/mm2

cebriggs7135 (approved):
Looks great.

ngehlenborg (approved):

PR #990, opened 2021-11-12
Added WGS directory structure schema

icaoberg:
The directory structure for the whole genome sequence assay type is nothing but a link to `fastq.yaml`. 

From my understanding, this should suffice for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.13 - in progress
+- Added WGS directory structure schema.
 - Cleaned up LC-MS directory structure schema.
 - Added links to examples in the portal for 5 assays.
 - Make LC fields optional.
--- a/docs/wgs/index.md
+++ b/docs/wgs/index.md
@@ -17,7 +17,7 @@ Related files:
 
 | pattern | required? | description |
 | --- | --- | --- |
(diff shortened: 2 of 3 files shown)


jswelling (commented):
You know the 'extras/.*' regex subsumes the 'extras/thumbnail.(png|jpg)' regex, right?  You don't actually need the second one?  If you want it there for stylistic reasons I'm OK with it.

icaoberg:
> You know the 'extras/.*' regex subsumes the 'extras/thumbnail.(png|jpg)' regex, right? You don't actually need the second one? If you want it there for stylistic reasons I'm OK with it.

@jswelling agree. Howver, the file `wgs.yaml` is a link to `fastq.yaml`. If we modify `fastq.yaml`, the change will affect at least 4 directory structure schemas.

What do you think I should do? Should I update `fastq.yaml`?

jswelling:
Wait, I just figured out that the 'extras' entry is hard-coded in schema_loader.py .  It's not coming from a .yaml file at all!  So no, no change is necessary.  I'll approve the merge as it stands.  Sorry for the confusion.

jswelling (approved):
Accepted- I hadn't understood that the redundant entry to accept thumbnails in extras was coming from code outside this PR.

PR #995, opened 2021-11-23
drop report in top of upload

### LINKED ISSUE #993: Dump validation report into extras
Add a CLI flag to dump the report into extras. Timestamp the file, so we have a record of successive validations, and consider adding a list of the files validated.

mccalluc:
Fix #993. Note that the same report is dropped in _each_ dataset directory. The `extras/` subdirectory seems like the best place to put it, and that means that it is being saved redundantly: There's no way to break up a report to get the parts that just pertain to a particular dataset.

These reports, in this location, will be visible in the portal. Not a good or bad thing, but no one should be surprised. 

Some things could be tweaked...
- Perhaps turn it on by default, so we could see if it was validated locally before being uploaded?
- It will always save the list in the flat list format. The nested format could be used instead?
- In practice, you may want to add the `--add_notes` flag as well, if you want more contextual information.


jswelling (commented):
When an Upload is split into individual Datasets,
- each Dataset is given an 'extras' directory if it does not already exist
- the content of the Upload's 'extras' directory (above the directory level of the Datasets) is copied redundantly into each new Dataset's 'extras' directory.
So maybe it's  not necessary to put the report in each individual 'extras' directory- placing it in a single top-level 'extras' directory would suffice wrt the downstream split-and-ingest process.

cebriggs7135 (approved):

mccalluc:
@jswelling 
> placing it in a single top-level 'extras' directory would suffice wrt the downstream split-and-ingest process.

Proposal:
- Put it at the top level,
- and require the user to specify an `ignore_glob`, rather than hard-coding an exception.

jswelling (approved):
Putting one copy in the top-level extras directory works for me.  I think ultimately the docs should suggest that validate_upload.py always be called with "--upload_ignore_globs extras", which is a little awkward.

PR #1007, opened 2021-12-02
Remove exceptions from the assay type match list; comment out assays

jswelling:


mccalluc (commented):
This seems to be a work in progress... Unless there's something particular I should look at, can you re-request review when it's complete? Thanks.

cebriggs7135:
@jswelling Can we zoom so you can explain this to me?

cebriggs7135 (commented):
@mccalluc What does the # you've added to some assays mean?

jswelling:
The problem with implementing this is that unknown assay names don't just appear in the list of assay names, they also have whole table and directory schemata associated with them.  The names can be commented out of the assay name list in enums.py but that causes a CI failure when the table schemata are scanned.  There is no easy way to "comment out" an entire schema.  The strategy of testing for consistency at CI time would work for the future if it could be made to work long enough to merge this PR but the path to doing so seems very long.

mccalluc:
Joel writes: 
> stuck because I can't get it to pass CI, because commenting out the problematic assays in the src/ingest_validation_tools/enums.py does not prevent testing of the corresponding table schemata during CI.  If I comment a problematic assay type out of enums.py the CI error caused by its presence goes away, but I introduce a new CI error downstream when it tries to generate the doc for the commented-out assay type.  I could use some advice on how to get this past CI.

mccalluc:
@jswelling: 
- The names that you're removing from the `assay_type` list in `enums.py`: Is your goal that they will no longer be allowed values in any metadata TSVs? I had thought that Nils had directed us to keep the assay names as they are here, and that any mapping between names would need to happen downstream.
- Setting that aside, where you've commented out `scATACseq` and `scRNAseq` in the schemas: Is there a reason you're focusing on just those two, or will you also want to remove all the other assay names that you comment out in `enums.py`?

mccalluc:
Superseded by #1022

PR #1010, opened 2021-12-03
deprecate contributors v0: require `is_contact`

### LINKED ISSUE #1003: Expire contributors-v0
Contributors v0 doesn't have the information that's required, so we should no longer accept uploads that use it.
Make PR, and Chris will decide when it can be merged, and if any additional communication is necessary.

mccalluc:
- Fix #1003

The IEC definitely does not want to receive any datasets where the contributors is missing `is_contact`... so we should deprecate v0 to make sure anything in the old form will be a validation failure.

When to do this, and whether any additional out-of-band communication is necessary, is for someone else to decide, but there should be a decision.


ngehlenborg:
Added to agenda for future portal team meeting.

cebriggs7135 (approved):

ngehlenborg (approved):

PR #1016, opened 2022-01-06
Add umi barcode fields to scrnaseq : generate V3

cebriggs7135:
The HuBMAP processing team requests that we add fields for umi position, size and read location for scRNAseq. umi values for Chromium V2 & V3 are known but these values should be available in the assay metadata for every scrnaseq assay type.


mccalluc:
Replaced by #1017. Closing.

PR #1017, opened 2022-01-07
Mccalluc/umi fields

mccalluc:
- Replaces #1016

The important bit:
```
$ diff src/ingest_validation_tools/table-schemas/assays/scrnaseq-v{2,3}.yaml --normal
2c2
< description_md: This schema is for single cell RNA sequencing (scRNAseq).
---
> description_md: This schema is for single cell RNA sequencing (scRNAseq). v3 adds `umi_*` fields.
9c9
<       - '2'
---
>       - '3'
72a73,80
> - name: umi_read
>   description: Which read file(s) contains the UMI (unique molecular identifier) barcode.
> - name: umi_offset
>   description: Position(s) in the read at which the umi barcode starts.
>   type: integer
> - name: umi_size
>   description: Length of the umi barcode in base pairs.
>   type: integer
```
Filing as draft for now. Questions to resolve:
- You annotated `umi_offset` as an integer, but you also say it describes  "Position(s)"... This suggests that what you actually want is a comma delimited list of (positive?) integers. Is that correct?
- For `umi_read`, is it a list of filenames, or a list of paths? If filenames, we could have a pattern constraint that forbids `/`... If paths, I don't think there is anything I can do without adding specialized code.
- For other measurements, we've had paired value and units fields... but base pairs seems obvious enough that I don't think that's necessary. Confirm?
- Can you give a date on which v2 (and for that matter v1 and v0) will no longer be accepted?
- HCA-scrnaseq began as a copy of this schema: Should these fields also be added there?


mccalluc:
> -For umi_read, is it a list of filenames, or a list of paths? ...

Ah: The directory schema here is just `[^/]+\.fastq\.gz`: Files should be at the top level, so I think it does make sense to add a pattern constrain to disallow slashes.

cebriggs7135 (approved):

mccalluc:
Chris writes:
> Dinh from UCSD just answered the question: “No, the UMI field does not need to have multiple values.”  I wonder why Matt thinks it should.

cebriggs7135:
Matt assumed UMI fields would sometimes require muptple input but Dinh (UCSD) says that's not the case for UMI barcode.

PR #1018, opened 2022-01-07
Mccalluc/field summary grid

### LINKED ISSUE #1015: Human readable report on which fields are used in which assays
As teams are developing new metadata schemas, they want to compare against existing field definitions so that terms can be reused where possible. Right now there's a spreadsheet that Chris maintains, but it would be less work, more accurate, and more up to date if this were part of the doc build process.

@cebriggs7135 :
- First column is the field name, second is the description, and following are boolean columns, one per assay type. Does that sound right? 
- Would any other information be useful, like the values for enums?
- Ordering: Alphabetically sort the rows and columns?
- What format(s) should be generated? TSV / HTML / Excel?
- With this in place would you be able to drop the spreadsheet, or are there other usecases that this doesn't cover?

mccalluc:
- @cebriggs7135 -- download the excel file and see if can replace the sheet that you've been maintaining... If not, let me know what the gaps are.
- @john-conroy -- Code review.

Fix #1015


cebriggs7135 (approved):

tests/test-generate-docs.sh
```diff
@@ -5,22 +5,45 @@ die() { set +v; echo "$*" 1>&2 ; sleep 1; exit 1; }
...
   diff -r $REAL_DEST $TEST_DEST \
-    || die "Update needed: $REAL_CMD; $LOOP"
+    || (
+      for ATTR in $ATTR_LIST; do
+        REAL_DEST="docs/field-${PLURAL}.yaml"
+        REAL_CMD="src/generate_field_yaml.py --attr $ATTR > $REAL_DEST"
+        echo $REAL_CMD
+      done
+      die "Update needed: Easiest to update all files at once."
+    )
```
john-conroy, line 29: A little unclear reading the diff. What's the purpose of the nested loop over `ATTR_LIST`?

john-conroy (approved):

tests/test-generate-docs.sh
mccalluc: If one of these files needs updating (perhaps there's a new field) they usually all need updating... So we dump all the commands to the console so the user can easily copy-paste and run the whole batch.

Let me add a comment.

tests/test-generate-docs.sh
```diff
@@ -5,22 +5,45 @@ die() { set +v; echo "$*" 1>&2 ; sleep 1; exit 1; }
...
 
   REAL_CMD="src/generate_field_yaml.py --attr $ATTR > $REAL_DEST"
   TEST_CMD="src/generate_field_yaml.py --attr $ATTR > $TEST_DEST"
 
   mkdir docs-test || echo "Already exists"
   eval $TEST_CMD || die "Command failed: $TEST_CMD"
   diff -r $REAL_DEST $TEST_DEST \
-    || die "Update needed: $REAL_CMD; $LOOP"
+    || (
+      for ATTR in $ATTR_LIST; do
```
mccalluc, line 23:
```suggestion
      # If one fails, often they all will, so it's easiest for the user to rerun them all:
      # The output from the lines below can be copy-and-pasted.
      for ATTR in $ATTR_LIST; do
```

PR #1022, opened 2022-01-14
Mccalluc/unblock joels assay work

mccalluc:
Alternative to @jswelling's PR #1007. If this is acceptable, before merging
- close that PR
- file an issue describing the follow-up work
- reference that new issue in the TODOs here.

(CI failed because uniprot service was unavailable: passed locally.)


jswelling (approved):
This does just what is needed.

PR #1024, opened 2022-01-14
Remove mention of thumbnail

mccalluc:
The mechanism that was proposed here was never actually used to my knowledge. 

[Thumbnails are available](https://github.com/hubmapconsortium/portal-ui/issues/2275), they just don't take this path. One example: [no extras here](https://app.globus.org/file-manager?origin_id=af603d86-eab9-4eec-bb1d-9d26556741bb&origin_path=%2F2c3540e2b1040c06612698d80ecab56a%2F). This will not affect validation: `extras/` is wildcarded.

Will require doc regen, but I wanted to get the proposal out there.


cebriggs7135 (approved):
OK

PR #1027, opened 2022-01-20
MIBI-add acquisition instrument vendor/model

cebriggs7135:
Added acquisition instrument vendor and model to the level 2 metadata for MIBI schema.
@ngehlenborg These 2 fields are required for all assays but they were omitted here so I'm adding them as we finish up the docs for MIBI. These should actually be included in level 1 metadata.


src/ingest_validation_tools/table-schemas/assays/mibi-v1.yaml
```diff
@@ -28,7 +28,26 @@ fields:
...
+- name: acquisition_instrument_vendor
+  description: An acquisition instrument is the device that contains the signal detection
+    hardware and signal processing software. Assays generate signals such as light
+    of various intensities or color or signals representing the molecular mass.
+  heading: Unique to this type
+  constraints:
+    required: true
+  custom_constraints:
+    forbid_na: true
+    sequence_limit: 3
```
mccalluc, line 40:
```suggestion
```
Defaults do not need to be given

src/ingest_validation_tools/table-schemas/assays/mibi-v1.yaml
```diff
@@ -28,7 +28,26 @@ fields:
...
+- name: acquisition_instrument_model
+  description: Manufacturers of an acquisition instrument may offer various versions
+    (models) of that instrument with different features or sensitivities. Differences
+    in features or sensitivities may be relevant to processing or interpretation of
+    the data.
+  constraints:
+    required: true
+  custom_constraints:
+    forbid_na: true
+    sequence_limit: 3
```
mccalluc, line 50:
```suggestion
```

mccalluc (changes requested):
- Please remove defaults
- Will wait for confirmation from Nils that this is ok to modify in place

ngehlenborg (approved):
I approve assuming  "modify in place" = "modify without bumping the version"

mccalluc:
I generate the docs here. Thanks.

mccalluc:
@cebriggs7135 -- Could you check the placement of these fields in the list? I believe that for other assays they are at the top of the "Unique to this type" list... but that's not the case here. Should they be?

cebriggs7135:
@mccalluc You are correct. Acquisition instrument fields should be at the top of the "Unique to this type" list.

mccalluc:
@cebriggs7135 -- Ok, good to know: They were misplaced at the start of this PR. If you can review the output now and confirm that the field order is what you want, I'd appreciate it. Will wait to merge until I get your confirmation.

cebriggs7135 (commented):
looks correct. thanks.

mccalluc:
Converting back to draft. This is not ready to go. The changes in https://github.com/hubmapconsortium/ingest-validation-tools/pull/1034 should be merged into this.

mccalluc:
Close in favor of #1035

PR #1031, opened 2022-01-21
Mccalluc/directory schema versions

mccalluc:
Earlier PR #928 was closed when it was asserted that TMCs could be compelled to stick with the same directory schema forever. That may not be viable. Coming out of the meeting this morning, @pdblood  directed me to resurrect this branch in case it is needed, perhaps for #1026. Keep as draft until we have a case where it is actually needed.


mccalluc:
Looking at the old code, this uses a heuristic to pick the appropriate schema: [pick the one with the fewest validation errors](https://github.com/hubmapconsortium/ingest-validation-tools/pull/1031/files#diff-4eac397eb6d8f2e4b35ef29e5c2d37f6c26bf44218e9bdb76ff288c128ff3739R40). It feels like this would be more confusing than having an explicit statement of which directory version this is. Try to figure out why this choice was made, and if it still makes sense.

jswelling:
@mccalluc my recollection re. using a heuristic is that I argued for it because it reduces the TMC's effort. I gather the alternative is to add an integer field to the metadata.tsv (in addition to the field specifying the metadata schema version).  If we can determine the correct directory schema without adding that field (and mixing the metadata and directory semantics), shouldn't we do so?  The downstream workflows will need to continue to support older directory schemata so that datasets can be re-run.

mccalluc:
@jswelling :
> my recollection re. using a heuristic is that I argued for it because it reduces the TMC's effort.

I don't believe this will be true. We know that it can be difficult for TMCs to understand error messages. If the software is guessing about which version the user intended, and sometimes getting it wrong, that's going to make the errors more confusing.

> I gather the alternative is to add an integer field to the metadata.tsv

No: I would like the datasets after version 0 (where we are now) to include a zero-length file with a name like `v123-codex` or whatever. As you rightly remark, we don't want to muddy the waters by confusing metadata schema versions and directory schema versions.

mccalluc:
Decision made in this meeting: Explicit: Add file to dataset directory whose name indicates version at the top-level.

mccalluc:
- Make a branch from this branch
- Rebase against main
- Add `dataset.json` to a new directory schema version for CODEX
- Merge this to main
- Close Ivan's PR (#1063), referencing the new PR from the branch of this PR

mccalluc:
- Rebase doesn't do what I want: New idea: Fresh branch from main, and then squash this into that.

PR #1034, opened 2022-01-25
mibi: add Ab path and description.md text

cebriggs7135:
added an antibodies_path field to the schema & text to the description.md. @ngehlenborg I believe this can also be added while keeping the same version.


docs/mibi/v1.yaml
```diff
@@ -1,5 +1,8 @@
 # Generated YAML: PRs should not start here!
```
mccalluc, line 1: Please note the protocol for submitting PRs.

mccalluc (changes requested):

docs/mibi/v1.yaml
mccalluc: I'll take your changes and implement them, but anything you can do to follow the protocol would help me.

mccalluc:
Close in favor of #1035

PR #1036, opened 2022-01-26
Mccalluc/move to level 1

### LINKED ISSUE #1033: move acquisition_instrument_vendor & _model fields to level 1 assay metadata schema
These fields are required for every HuBMAP assay so they belong with level 1 so they don't inadvertently get left out of new schema.

mccalluc:
Fix #1033. It might be good to review the earlier documentation and see if there was a reason the acquisition instruments weren't in the top list to being with.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.14 - in progress
+- Move "Unique to this type" below the acquisition instrument fields.
 - Add acquisition instrument fields to MIBI that were left out by mistake.
 - Just use pytest to run doctests
 - Headers are no longer properties of fields.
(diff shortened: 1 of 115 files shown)


cebriggs7135:
Based on this doc from 2020, it appears that we envisioned acquisition as something only relevant to imaging assays; https://docs.google.com/document/d/1g82GpCpFDKew60XzAO4Siaw3ZXJjwsaCpgPwhqQZxIY/edit#heading=h.xyl7oa7lr0a2
"Level 2 Attributes Section (here: “imaging”)
Different categories of assays (imaging, mass spectrometry, sequencing) will use different sets of Level 2 attributes. For this preview, we only cover those attributes shared by all imaging assays.
acquisition_instrument_vendor
Description: An acquisition_instrument is the device that contains the signal detection hardware and signal processing software. Assays generate signals such as light of various intensities or color or signals representing molecular mass.
Allowed Values: Akoya, Zeiss,10xGenomics, ..."

ngehlenborg (approved):

cebriggs7135 (approved):

mccalluc:
Tests are failing because portal is in maintenance mode, so `https://portal.hubmapconsortium.org/browse/dataset/bad` is returning 200 instead of 404. Wait until server maintenance is done, then re-run tests.

Issue #1037, opened 2022-01-27
Revert directory structure changes


mccalluc:
From Joel:
> we need to resolve a pair of problems affecting the CellDIVE datasets, which I discovered in the process of re-running a CellDIVE test case on DEV.  ingest-validation-tools has new commits since the CellDIVE data was accepted which invalidate the metadata.tsv files for that data.  Specifically, (1) the contributors.tsv file matches contributors-v0, and that has been deprecated, and (2) code has been added (why???) to reject directory names ending in / .
>
> It's easy for me to work around this for my test on DEV, just by resetting ingest-validation-tools to its state before these changes were made, but I can't do so on PROD without losing updates for new assays that we need to keep.  So I'd like you folks to back out those two changes until after the CellDIVE data has been processed and published. (edited) 
>
> Even then, the deprecation of the contributors.tsv file will be a show-stopper if we ever need to reprocess this data.  Is there some option I can use with ingest-validation-tools to omit deprecation errors?

Possible path forward:
- [ ] CLI option to allow validation against deprecated directory schema.
- [ ] Find code change relating to paths ending with `/`, and decide next step.

mccalluc:
@jswelling 
- See if #1038 will be useful.
- Looking back, disallowing trailing slashes came in with #1002. That PR also provides a more clear error message if an unexpected file or directory is present in the submission, but I think the slash checking can be separated from that... but there may have been a motivation that escapes me now.

mccalluc:
Proposed PRs have been merged, resolving this issue.

PR #1038, opened 2022-01-27
ignore deprecation flag

mccalluc:
Towards #1037, I think.

@jswelling : If you're checking out an older version as a hack, then even without this being merged to master, it might be useful to you, if you'd like to check out this version in airflow. Can you explain how you'll use this option? 
- Will it be on all the time, or will you just turn it on sporadically?
- Would it be useful to make it more fine grained, to only allow deprecated versions for some schemas? Or would that be gratuitous?
- Would it be useful to emit some kind of warning, even if the flag is set? (I haven't been excited about implementing warnings that aren't failures... adds complexity to the code / makes validation criteria less clear-cut.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v0.0.14 - in progress
+- Add CLI option to allow the use of deprecated schemas.
 - Add acquisition instrument fields to MIBI that were left out by mistake.
 - Just use pytest to run doctests
 - Headers are no longer properties of fields.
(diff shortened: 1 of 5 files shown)


cebriggs7135 (approved):

mccalluc:
Asked Joel for clarification on Slack: 
> can you clarify how/when it will be used? Will you keep it on by default, or just turn it on when there are older datasets to reprocess?

mccalluc:
Tests are failing because portal is in maintenance mode, so `https://portal.hubmapconsortium.org/browse/dataset/bad` is returning 200 instead of 404. Wait until server maintenance is done, then re-run tests.

mccalluc:
Joel clarifies:
> Each dataset gets validated multiple times, and with the current workflow the last chance for a contributor to fix a dataset comes when an Upload is Submitted. My thinking was to set this flag to true for any check that comes after that validation, when we would already have “officially” accepted the dataset.

I'll update the flowchart.

PR #1041, opened 2022-02-03
Add snRNAseq_10xGenomics_v2 to scrnaseq assay schemas

jswelling:
This PR adds the assay type snRNAseq_10xGenomic_v2 to all the scrnaseq table schemas.


cebriggs7135 (approved):

src/ingest_validation_tools/enums.py
```diff
@@ -57,10 +57,11 @@
         # 'TMT (proteomics)',
         'WGS',
         'SNARE2-RNAseq',
-        'snRNAseq',
+        'snRNAseq',  # equivalent to snRNAseq-10xGenomics-v3
         'scRNAseq-10xGenomics',  # Only needed for scrnaseq-v0.yaml.
         'scRNAseq-10xGenomics-v2',
         'scRNAseq-10xGenomics-v3',
+        'snRNAseq-10xGenomics-v2',
```
mccalluc, line 64: This confuses me: So the new value that's being added, `snRNAseq-10xGenomics-v2` is actually for an older version of the assay than was supported previously?

mccalluc (commented):
I'm confused by this. Besides the history of this particular assay, how will labs know what this means, and how to apply it? Is there some external documentation that explains to data providers what the different versions mean? Downstream, will this distinction show up in the Portal? How will users there know what it means?

src/ingest_validation_tools/enums.py
jswelling: @mccalluc snRNAseq already maps to the description string 'snRNA-seq (10x Genomics v3)'; 106 datasets already exist for that assay.  The -v3 part refers to the length of an internal primer, referred to as 'chromium-v3' in the literature. What's happened here is that we have encountered a 'chromium-2' version of that assay.  To describe it in the metadata.tsv, we introduce the assay type string 'snRNAseq-10xGenomics-v2', which already exists on the search-api side and has a description string 'snRNAseq (10x Genomics v2)'.  The distinction between -v2 and -v3 is not covered in the docs, and in fact neither is the distinction between 'sn' and 'sc' (single-cell vs single-nucleus), but both of those qualifiers have been present on the scRNAseq collection of assays for some time.  I'd argue that we at least are not making the lack of documentation any worse by adding this case.

cebriggs7135: @jswelling  Duly noted.I'll put this on Slack as a reminder to add a line or 2 to the scnraseq documentation after the new host site is up and running.

mccalluc: @jswelling -- Thanks for the explanation. 

@cebriggs7135 -- I think this would be good to explain that in the docs here, too. I'll file. Can you explain what you mean by putting a reminder on Slack? In general, I think github issues are a good place for our TODO lists.

mccalluc (approved):
Thanks!

PR #1058, opened 2022-03-21
New sample metadata schemas

### LINKED ISSUE #1057: Implement new sample schemas
https://docs.google.com/spreadsheets/d/1yqgPaVsUNpEavZOgl0bZsiTBABuVXJtq/edit#gid=233528636

mccalluc:
- Fix #1057: [spec](https://docs.google.com/spreadsheets/d/1yqgPaVsUNpEavZOgl0bZsiTBABuVXJtq/edit#gid=233528636) / [q+a](https://docs.google.com/document/d/1jTlA0zO73j0djjQV03wpsgff4B05he3bipoTEz46ae0/edit)

Notes:
- For several enums, `None` is specified as one of the allowed values. An alternative would be to mark these fields as optional, but if that were done, we wouldn't know if the value were actually `None`, or if the lab just hadn't bothered to fill it in.
- Currently no validation on Sample ID. I think there are differences of opinion on this, but please do not try to use comments here to sort this out. **Filing as draft until resolved.**
- Version in spec is `2`... but on other schemas, our first explicit version number is `1`, so I've done that here as well.
- `notes` does not have a description in the spec, so I have just used "Notes". 
- `suspension_enriched` is spec'ed as an enum, `yes`/`no`, rather than a boolean. I have implemented the spec.
- Description for `suspension_enriched_target` seems vague to me... but I'm not a lab scientist: "If the suspension was enriched, then this is the target of the enrichment."


ngehlenborg (commented):
In principle, I approve (pending resolution of the sample ID question) but I think it would be better to switch the `suspension_enriched` field to a Boolean field. Are we generally using Booleans instead of "yes" and "no"?

mccalluc:
Nils gives approval for status quo on slack. Will resolve conflicts and merge.

PR #1063, opened 2022-03-29
CODEX directory structure schema update

icaoberg:
Added `dataset.json` file to schema.

PR #1067, opened 2022-03-31
Dir schema versions + new CODEX schema

### LINKED ISSUE #1061: Change CODEX directory schema: add new metadata JSON file
- Does this change have buy-in from other groups providing the same kind of data? Yes
- Can you quote or reference any outside documentation that motivates this change? Yes
- Is this change backward compatible, or does it make validation more stringent? Yes

The updated CODEX metadata should be in a different file, `dataset.json`. Add this to the schema as a new required file.

@icaoberg will evaluate this request, and if it can move forward,
will file a PR for @pdblood to review,
and @mccalluc will make sure tests pass before merging.

mccalluc:
In progress:
- Starting from a squash of #1031
- Need to clear out cruft that is not related to dir schema versioning. (I think it came in with improperly resolved merges, but I'm not sure.)
- Need to make version schema explicit, as agreed.
- Need to add new `dataset.json` to CODEX: replaces #1063; Will fix #1061.


icaoberg (approved):
@mccalluc this looks good to me and seems to implement what we discussed so far.

I will try to locally validate some directories after the merge.

PR #1073, opened 2022-04-18
After failure, explain how to continue from last error

mccalluc:
This functionality has been in place for a while, but it hasn't been communicated.


jswelling (approved):
Definitely useful, and something I did not know.

PR #1074, opened 2022-04-18
type checks

mccalluc:
Branched from the directory schema PR. Coming from that experience, I thought that adding typing would make the code more maintainable: Even in just the first few files, I've caught places where the parameters names no longer reflect the data they contain.


jswelling (approved):
I am definitely in favor.

PR #1083, opened 2022-05-04
re-enable enum validation

mccalluc:
The only ones that remain commented out are
```
        # 'SPLiT-Seq',
        # 'TMT (proteomics)',
```
Do you know what's up there?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 ## v0.0.14 - in progress
+- Turn validation of enums back on.
 - Fix rendering bug on CODEX page by adding linebreaks.
 - Add type hints.
 - Implement versioning for directory schemas.
--- a/src/ingest_validation_tools/enums.py
+++ b/src/ingest_validation_tools/enums.py
@@ -14,11 +14,10 @@ from typing import Dict, List
 >>> print('\\n'.join(remote_mismatch))
 <BLANKLINE>
 '''
-# The list above should be empty: That means that all the assays
+# TODO: The list above should be empty: That means that all the assays
 # listed below are recognized by the assay service.
-# TODO: Joel will progressively uncomment names below, and when that is done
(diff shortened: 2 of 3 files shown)


jswelling (changes requested):
bulk RNA, NanoDESI, NanoPOTS, and the MS/LC-MS no-modifier/top-down/bottom-up set should be fine.  I believe TMT is going away completely.

mccalluc:
@jswelling -- It's not actually a problem that some of these names are no longer valid, in fact, that's what I'd expect. I've added a comment that tries to explain what this list is about:
```
# The list above should be empty: There should be no mismatches;
# ie: All the assays listed below are recognized by the assay service.
#
# The assay_type list is *all* the values which have ever been used,
# including schemas which are currently deprecated.
# Each schema lists the particular values which are valid for it.
# This redundant list catches mistakes like a typo being introduced
# when an existing schema is copied, pasted, and tweaked to make a new version.
```

mccalluc:
Pinged Joel on Slack. Need to add CODEX2 to the list.

jswelling:
Hm; I had not actually considered that some assays in enums.py might remain present after they are deprecated, but that makes sense.  What is the policy wrt assay names that are actually abandoned, like 'LC-MS (untargeted)'?  The plan as I understand it is to modify existing assays on disk and in Neo4J such that that assay type is never seen again, so a metadata.tsv file containing that string would appropriately be called 'Invalid'.  
The following assay types have no corresponding entry in search-api: 'CE-MS', 'DESI', 'SIMS-IMS'.  They could easily be added, though I don't know that we have any providers for those assay types.  My understanding is that 'MS (shotgun-lipidomics)' and 'LC-MS (metabolomics)' are now invalid names.  The problem arises when one of the deprecated old names does not map uniquely to a single new 'approved' name.  If a metadata.tsv file contains the old name and is considered valid, the correct new name is indeterminate.

jswelling (approved):

PR #1090, opened 2022-05-19
Mccalluc/report directory schema version

mccalluc:
- Fix #1087
- Review and merge #1091 first, hopefully (Work is duplicated here: Easier to handle that question in isolation.)

@jswelling:
- There's not an explicit test at the python level for this new info, but the CLI output fixtures (in yaml) show you what to expect.
- Currently, it would be possible for different datasets under a single TSV to use different directory schema versions... so right now this returns a list of version numbers, rather than a single. Would you prefer that such a situation be an error, and only one directory schema version is allowed across all datasets referenced in a TSV?

@john-conroy -- Code review / Good to have some familiarity with this machinery.

PR #1091, opened 2022-05-23
Mccalluc/remove html opts

mccalluc:
Back at the start of this, I had thought that structuring the error list hierarchically would be readable, and an html structure with collapsible sections would be the easiest way to review this kind of structure... But Bill suggested that what folks really want is just a flat, human readable list, and my sense is that that's the format that gets used.

This PR removes the HTML generation code; The motivation is:
- #1090

With the output changing, a bit of work would be required here, so seems like a good time to simplify... but if it turns out that someone actually needs/wants this, it wouldn't be hard to bring it back, and do the additional work that's needed. We don't have a user-group email list, and even if we did, I think I would lean towards asking for forgiveness rather than permission... but I could be making the wrong call.

If this is approved, I would plan to merge this into main, and then merge than into #1090.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 ## v0.0.14 - in progress
+- Remove HTML reporting options.
 - Updated IMS directory structure schema.
 - Add Clinical Imaging schemas.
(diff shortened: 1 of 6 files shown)


jswelling (approved):
Looks good to me.

PR #1092, opened 2022-05-23
Reactivate enums.py doctest; uncomment 6 MS assays

jswelling:
This PR reactivates the doctest in enums.py, and uncomments the following assays: 'MS Bottom-Up', 'MS Top-Down', 'LC-MS Top-Down', 'LC-MS', 'LC-MS Bottom-Up', 'MS'.  It requires https://github.com/hubmapconsortium/search-api/pull/496 , and will not pass CI until that PR has been deployed.

### DIFF
--- a/src/ingest_validation_tools/enums.py
+++ b/src/ingest_validation_tools/enums.py
@@ -1,25 +1,5 @@
 from typing import Dict, List
 
-
-'''
->>> import requests
->>> local_names = shared_enums['assay_type']
->>> remote_mismatch = []
->>> for name in local_names:
-...     response = requests.post(
-...         'https://search.api.hubmapconsortium.org/assayname',
-...         json={'name':name}, headers={'Content-Type': 'application/json'}).json()
-...     if 'error' in response:
-...         remote_mismatch.append(name)
->>> print('\\n'.join(remote_mismatch))
-<BLANKLINE>
-'''
-# The list above should be empty: That means that all the assays
-# listed below are recognized by the assay service.
-# TODO: Joel will progressively uncomment names below, and when that is done
-#       _validate_level_1_enum() can be renabled in schema_loader.py.
-#       https://github.com/hubmapconsortium/ingest-validation-tools/issues/1023
-
 shared_enums: Dict[str, List[str]] = {
     'assay_category': [
         'imaging',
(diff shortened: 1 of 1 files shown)


src/ingest_validation_tools/enums.py
```diff
@@ -87,3 +67,27 @@
...
+    """
+    # The list above should be empty: That means that all the assays
+    # listed below are recognized by the assay service.
+    # TODO: Joel will progressively uncomment names below, and when that is done
+    #       _validate_level_1_enum() can be renabled in schema_loader.py.
+    #       https://github.com/hubmapconsortium/ingest-validation-tools/issues/1023
+    pass
+
+if __name__ == "__main__":
+    main()
```
mccalluc, line 93: There is no `main` defined.

src/ingest_validation_tools/enums.py
```diff
@@ -87,3 +67,27 @@
         'phosphopeptides'
     ]
 }
+
+def _test_assay_name_compatibility():
```
mccalluc, line 71: This function is a no-op and is never called... were the top-level doc tests not being called, or was there some other problem?

mccalluc (changes requested):
This seems to overlap with #1083.

icaoberg (changes requested):
I see your uncommented list however I don't understand why you are calling `main` but there is no main function.

Maybe I am missing something.

src/ingest_validation_tools/enums.py
jswelling: Ah, true- I failed to remove a development test. Fixed.

icaoberg (approved):
Awesome.

src/ingest_validation_tools/enums.py
jswelling: The top-level doctest was not being called; I had to bind it into a function to get 'pytest --doctest-modules' to find it.  It may be that your change to the TODO block in the comment someone reactivated the doctest in your corresponding PR, but I don't understand the mechanism that would have done that.

jswelling:
All changes this was intended to bring to the code base have been handled by https://github.com/hubmapconsortium/ingest-validation-tools/pull/1083

PR #1102, opened 2022-05-24
Add "CODEX2"

### LINKED ISSUE #1095: Change CODEX metadata schema: allow 'CODEX2' assay_type in addition to 'CODEX'.
@ngehlenborg @mccalluc: This change is to allow processing of the new format of CODEX (which will be designated assay_type CODEX2) while we continue to process the old format of CODEX (which will continue to use the assay_type CODEX).

mccalluc:
- Fix #1095; Requested by @pdblood 

I am not confident that this in the right course, but I shouldn't be the blocker.
- If the structure of the submission is changing, should that be a new directory schema version rather than a change to the metadata schema?
- Is this the best field to change? `preparation_instrument_vendor` is also available.
- Should this be applied to all versions of the codex metadata schema, or only the most recent?
- Is there documentation that can explain to future users of this data how CODEX and CODEX2 differ?


jswelling:
The assays we need to call assay_type CODEX2 are precisely those that are now called CODEX with directory schema version 1.  A more desirable solution would be to rewrite the downstream assay management software so that the workflow used could be distinguished based on not just the assay_type string but the directory schema version.  The first changes to ingest-validation-tools needed to do this are already in process, with directory schema version becoming available at validation time.  Unfortunately the coding necessary to implement the downstream parts of this cannot happen before the end-of-June deadline to have ingested and processed these datasets.  This route, introducing a new assay_type, can happen before the deadline.  The hope is that we can unwind it again after the deadline when the downstream changes have made it through the development pipeline.

ngehlenborg:
@pdblood Handing this over to you and your team for now.

pdblood:
Thanks for the input. As Joel says, although not ideal, this is the best solution to be able to ingest both Vanderbilt CODEX (new format) Stanford/UFL CODEX now, which we must do for these awards ending in June.

However, the good news is that we know exactly what CODEX data is left to process in the old formats, and all new CODEX data will be ingested in the new format. Once all legacy format data is ingested (in the next month or so) we'll immediately revert the change, and update the metadata, which will be easy to do for the limited amount of new format CODEX data that we need to ingest in the near term.

Going forward we'll get the better solution Joel outlines in place so we can better manage harmonization of assay formats in the future.

pdblood (approved):

PR #1106, opened 2022-06-03
Add the plugin directory to sys.path before import

jswelling:
Import statements within plugin validators were found to fail.  This PR fixes this problem by temporarily prepending the plugin validator directory to sys.path while the validators are being loaded.


src/ingest_validation_tools/plugin_validator.py
```diff
@@ -130,6 +131,7 @@ def validation_error_iter(base_dir: PathOrStr, assay_type: str,
         for name, obj in inspect.getmembers(mod):
             if inspect.isclass(obj) and obj != Validator and issubclass(obj, Validator):
                 sort_me.append((obj.cost, obj.description, obj))
+    sys.path.pop(0)
```
mccalluc, line 134: This worries me a little... imaging that the insert on 119 is removed at some point, but we forget that this is paired. Prefer 
- `with` context, 
- or maybe just check if `plugin_dir` is already in the list, and only add if not?

jswelling: That's a fair concern.  Thank you; I found an appropriate idiom and will update the PR.

mccalluc (approved):

PR #1110, opened 2022-06-23
Reminder after TSV validation

### LINKED ISSUE #1109: If just a TSV validation was run, warn that the submission still needs validation
Joel:
> ... The metadata.tsv file had individual data files (fastq.gz files) in the data_path column rather than directory names.  Is there no way to detect this during local validation? ...

Chuck:
> I would say that folks should be validating the entire submission locally. Is there a reason that’s not being done?

J:
> Well, it clearly wasn’t done in this case, even though the user thought they had followed all the documented steps.

C: 
> I could add a warning to the TSV validation, if it passes, explaining that this did not validate the entire submission?

J:
> That might be helpful to us in diagnosing problems like this one.

mccalluc:
Fix #1109 

@jswelling -- This is trivial, but it feels like a software fix to a communication problem. Before approving, can you either track down how the misunderstanding arose, or give me contact info for the TMC and the curator so I can? Maybe the README here should be more clear, or maybe there is other documentation being referenced?

Also: Is the curator on your side actually taking on the work described in the flow chart? If not, should the flow chart change, or should they be picking up more of the work?


sunset666 (approved):
Small changes that has no impact on the project.

mccalluc:
@bhonick confirms on slack that he will follow-up on the documentation:

> I can confirm I have a tasking index entry for this on my end. This dovetails with what is being planned for validation education/documentation with SenNet

PR #1121, opened 2022-07-12
Use the assay service to describe assays

### LINKED ISSUE #1118: Include Assay service name in generated docs

mccalluc:
- Fix #1118
- @ngehlenborg 
  - I think you wanted more clarity on how the assay names here relate to those in the portal. This attempts to provide that: Is it enough?
- @jswelling 
  -  A number of the names here aren't in the assay service... maybe we just haven't had any submissions of those, and so you haven't needed to make updates there?
  - Hitting the assay service at doc generation time does slow down the process a bit... is it worth having a local cache... or will that just add complexity?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 ## v0.0.15 - in progress
+- Use the assay service to describe how assays are represented in the Portal.
 
 ## v0.0.14 - 2022-06-23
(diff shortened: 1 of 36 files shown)


ngehlenborg (approved):
Extremely helpful!

sunset666 (approved):
It has the potential of slow processing due to individual queries to search-api.

PR #1150, opened 2022-11-11
Updated CODEX v0

icaoberg:
Updates regex so that the `segmentation` file is not expected to be found in the `extras` folder.

### DIFF
--- a/src/ingest_validation_tools/directory-schemas/codex-v0.yaml
+++ b/src/ingest_validation_tools/directory-schemas/codex-v0.yaml
@@ -12,7 +12,7 @@ files:
     is_qa_qc: True
     required: False
   - 
-    pattern: '(raw|processed)/config\.txt|(src_[^/]*|drv_[^/]*|extras)/[sS]egmentation\.json' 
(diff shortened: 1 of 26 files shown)


bhonick (approved):

src/ingest_validation_tools/directory-schemas/codex-v0.yaml
```diff
@@ -12,7 +12,7 @@ files:
     is_qa_qc: True
     required: False
   - 
-    pattern: '(raw|processed)/config\.txt|(src_[^/]*|drv_[^/]*|extras)/[sS]egmentation\.json' 
+    pattern: '(raw|processed)/config\.txt|(src_[^/]*|drv_[^/]*)/[sS]egmentation\.json'
```
sunset666, line 15: We still have some codex data uploads that uses the raw folder for this, not sure if that is the case anymore or if it should be allowed.

icaoberg: @sunset666 this is what @pdblood asked me to do. @bhonick can you confirm this is correct? cc @mruffalo

PR #1161, opened 2022-12-20
Change field type of `preparation_maldi_matrix` to a string rather than restrictive enum

### LINKED ISSUE #1160: Change Imaging mass spectrometry metadata schema: some field types and required fields should be changed
- Does this change have buy-in from other groups providing the same kind of data?
Yes, there were mistakes made in the creation of the v2 of the imaging MS schema.

- Is this change backward compatible, or does it make validation more stringent?
Changes would be backward compatible. These changes affect which fields are required and some field types.

For instance, `preparation_maldi_matrix` field is set to be an enum with 3 potential values. There are hundreds of MALDI matrices, this field should be a string. 
https://hubmapconsortium.github.io/ingest-validation-tools/ims/#preparation_maldi_matrix

NHPatterson:
This fix changes this field in v2 of the metadata schema to be less restrictive. At some point this was conceived as an enum with 3 choices, none of which Vanderbilt uses. This change makes this an open string field. 

## alternatives
Expand the enum to a large list of common MALDI matrices. However, in the future someone could still submit data with a different or custom matrix and then the enum would need to be expanded again. 

Closes #1160.


jswelling (approved):

icaoberg:
@jswelling it didn't pass the CI test on 3.6. @sunset666 can you take a look at this?

NHPatterson:
> @jswelling it didn't pass the CI test on 3.6. @sunset666 can you take a look at this?

Not my place to make these decisions and certainly don't know all the implications but py 3.6. It reached end of life in Dec 2021: https://endoflife.date/python

Issue #1138, opened 2022-09-20
Change stained microscopy directory schema: Allow different file formats than .scn


NHPatterson:
- Does this change have buy-in from other groups providing the same kind of data?

This should expand the file types of raw data that can be uploaded. Right now, the specification only allows `.scn` files to be uploaded which are from Vanderbilt TMC's early specification. We at Vanderbilt TMC now do this scanning with a Zeiss instrument that creates `.czi` files.  Although for older data we may still have `.scn` files so I don't recommend a total deprecation of this extension.

- Can you quote or reference any outside documentation that motivates this change?

We'd like to upload our data to HuBMAP. 

- Is this change backward compatible, or does it make validation more stringent?

This should be backwards compatible if the acceptable file extensions are expanded.

@icaoberg will evaluate this request, and if it can move forward,
will file a PR for @pdblood to review,
and @mccalluc will make sure tests pass before merging.

NHPatterson:
## Changing direction to be more general

After speaking with @pdblood, we realized this issue may have broader implications. 

- The directory schema is listed as "PAS Stained Microscopy" but the metadata.tsv has a [stain](https://hubmapconsortium.github.io/ingest-validation-tools/stained/#stain) field where one can select PAS or H&E or another histological stain. Going forward, this should be generalized to "Stained Microscopy" and be an umbrella for H&E, PAS, and other histological stains that are usually imaged by brightfield microscopy. 

- The raw data should be less vendor focused. In version 1, the schema checks for `.scn` and we initially proposed to add `.czi` to accommodate us, Vanderbilt, specifically as we capture these images on a new instrument. A better idea would be to upload any raw data as a losslessly converted OME-TIFF as well, making the level 0 data more FAIR. This is aided specifically by the fact that there are now several easy to use OME-TIFF conversion tools.

## OME-TIFF conversion tools:

### Command-line
#### bioformats2raw -> raw2ometiff
Probably the tool of reference for this task, uses command line to perform conversion. 2-step process.
[bioformats2raw](https://github.com/glencoesoftware/bioformats2raw)
[raw2ometiff](https://github.com/glencoesoftware/bioformats2raw)

### GUI
#### QuPath
GUI interface, users can open images in QuPath and convert through file menu options.
[QuPath](https://qupath.github.io/)

#### Kheops
GUI interface through FIJI imageJ. 2-step install process through FIJI update sites.
[Kheops](https://github.com/BIOP/ijp-kheops)

[...]

PR #1164, opened 2023-01-03
Update directory schema for stained-v0.yaml for more raw file types

NHPatterson:
## background
Vanderbilt is the only lab that has produced this assay type and we initially helped build the directory schema but we have changed some instruments and no longer produce `.scn` files bur rather `.czi` files, hence needing to expand the schema definition.

## what is changed
The current directory schema expects a `.scn` file extension for the `raw_microscopy` file (https://github.com/hubmapconsortium/ingest-validation-tools/blob/9eb8f93f388b9094b0f4574265f628bd568cacbb/src/ingest_validation_tools/directory-schemas/stained-v0.yaml#L20). 

This PR would expand this to other file types (.czi, .tif, and .tiff). 

## HIVE considerations
None of the files in `raw_microscopy` sub-directory are used by HIVE processes, the pyramid visualized on the portal is generated from `processed_microscopy` sub-directory's OME-TIFF (which is what is desired). 

This is related to #1138 but that needs more elaboration to truly close the issue.


icaoberg (approved):
This is correct and aligned with @NHPatterson intentions.

PR #1169, opened 2023-01-24
Added MxIF directory structure schema v0

icaoberg:
Added new directory structure schema built by @ropelews and @j-uranic .

### DIFF
--- a/src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
+++ b/src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
@@ -2,14 +2,18 @@ files:
   # Level0/
   -
     pattern: '(Level0|level0)_raw\/cyc_[\d]+\.(tif|tiff)'
-    description: 'MXiF raw tile data by cycle and channel.'
+    description: 'MxiF raw tile data by cycle and channel.'
     required: True
     is_qa_qc: True
   -
-    pattern: '(Level0|level0)_lab_pre-processing\/cyc_[\d]+\/tile_layout\.json'
-    description: 'MXiF tile layout in pixel coordinates in x and y.'
+    pattern: '(Level0|level0)_raw\/cyc_[\d]+\/tile_layout\.json'
+    description: 'MxiF tile layout in pixel coordinates in x and y.'
     required: True
     is_qa_qc: False
+  -
+    pattern: '(Level0|level0)_raw\/dataset\.json'
+    description: 'Additional specific metadata.'
+    required: True
 
   # Level1/
   -
--- a/src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
+++ b/src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
@@ -1,5 +1,31 @@
 files:
+  # Level0/
   -
-    pattern: 'TODO'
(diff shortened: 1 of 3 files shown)


j-uranic (commented):
The updates match the file mxif_draft_updated.yaml.

src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
```diff
@@ -1,5 +1,31 @@
 files:
+  # Level0/
   -
-    pattern: 'TODO'
-    description: 'Directory structure not yet specified.'
-    required: True
\ No newline at end of file
+    pattern: '(Level0|level0)_raw\/cyc_[\d]+\.(tif|tiff)'
+    description: 'MXiF raw tile data by cycle and channel.'
```
NHPatterson, line 5: should be `MxIF` rather than `MXiF`.

src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
```diff
@@ -1,5 +1,31 @@
...
+    required: True
+    is_qa_qc: True
+  -
+    pattern: '(Level0|level0)_lab_pre-processing\/cyc_[\d]+\/tile_layout\.json'
+    description: 'MXiF tile layout in pixel coordinates in x and y.'
```
NHPatterson, line 10: should be MxIF rather than MXiF.

src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
```diff
@@ -1,5 +1,31 @@
...
-    pattern: 'TODO'
-    description: 'Directory structure not yet specified.'
-    required: True
\ No newline at end of file
+    pattern: '(Level0|level0)_raw\/cyc_[\d]+\.(tif|tiff)'
+    description: 'MXiF raw tile data by cycle and channel.'
+    required: True
+    is_qa_qc: True
+  -
+    pattern: '(Level0|level0)_lab_pre-processing\/cyc_[\d]+\/tile_layout\.json'
```
NHPatterson, line 9: This tile layout information is not lab pre-processing and is raw data. It tells us where each tile lives in the space of the full image prior to stitching. I think these files should live in the same directory as the tile data. I suggest (regex might not be perfect below):

pattern: '(Level0|level0)_raw\/cyc_[\d]+\tile_layout\.json'

This means that each tile layout information file is stored next to the tile images.

src/ingest_validation_tools/directory-schemas/mxif-v0.yaml
```diff
@@ -1,5 +1,31 @@
 files:
+  # Level0/
```
NHPatterson, line 2: A `dataset.json` file similar to CODEX is missing. This would essentially be a subset of CODEX's `dataset.json` because of the lack of compatibility between the two. This would have essential metadata like number of cycles, channel names, which channels are nuclear stains & membrane stains, etc. that HIVE tools use to process the data.

NHPatterson (changes requested):
I think this looks good but needs a few things added.

j-uranic (commented):
This looks complete.

PR #1187, opened 2023-04-21
J uranic/publication assay schemas

j-uranic:
Uploaded directory schema and table schema .yaml files for new Publication assay.


jswelling:
I have tested this successfully against the prototype Publication directory cff698cf72e09e712afd7b1905ce9dca on TEST.

jswelling (changes requested):
I think the regex element '/.+\.json' actually permits additional '/' elements and thus permits subdirectories before the filename, so the elaborated version at line 14 is unnecessary.

j-uranic:
> I think the regex element '/.+.json' actually permits additional '/' elements and thus permits subdirectories before the filename, so the elaborated version at line 14 is unnecessary.

(Vignettes|vignettes)/(Vignette|vignette)_\d+/**.+/**.+\.json'

I had included the .+/ (bolded above) to accommodate the possibility of a subdirectory in the example directory tree ("mysubdir" below). If it's not needed, it can be deleted.

```
- vignettes
- - vignette_01
- - - mysubdir
- - - - conf.json
```

ngehlenborg:
No subdirectories within the vignette directories are needed or should be allowed. Can we remove that option?

j-uranic:
> No subdirectories within the vignette directories are needed or should be allowed. Can we remove that option?

Sure, that mapping has been removed.

j-uranic:
Thanks for the clarification, I updated it to /[^/]+\.json and now it is only matching examples without a subdirectory.

jswelling (approved):

jswelling (approved):
My approval doesn't really count here, since a bunch of the changes are actually mine.

PR #1197, opened 2023-07-27
Jpuerto/cedar docs

jpuerto-psc:
This PR modifies the way that doc generation functions. This PR also adds new YAML files to distinguish between a "local" YAML file and a metadata schema that is hosted on CEDAR. This will allow us to point out towards the CEDAR OpenView pages for those schemas.

Open items:
- [ ] Can we point out towards Excel/TSV files hosted on CEDAR? This would allow us to avoid having to manually upload/maintain these files and integrate that into our CI/CD pipeline.


gesinaphillips (approved):

docs/field-assays.yaml
pdblood, line 1: Juan I clarified with Stephen that "deprecated" just means that we don't expect more data of those types to be submitted, so we are not updating those schemas to the next-gen schemas. I'm not sure "deprecated" really conveys that. Maybe just "No Plans to Update" would be better?

pdblood: MxIF is not deprecated either, just renamed.

pdblood (commented):
I made a couple of comments around assay types labeled "deprecated" that should be addressed either now or after initial merge.

docs/field-assays.yaml
jpuerto-psc: @pdblood - I've just removed those new schemas. They'll remain up and visible since we will continue to accept this data.

PR #1204, opened 2023-08-09
Juranic/histology schema

jpuerto-psc:
Adds histology schema and updates the extras directory description


j-uranic (approved):
Changes look good to me.

PR #1217, opened 2023-10-04
CEDAR tests

gesinaphillips:
Added manual tests for CEDAR templates. These are manual because I don't want them to get added to the GitHub Portal CI action because they require API calls. Most changes are due to the addition of the `/examples/dataset-examples/bad-cedar-*` and `/good-cedar-*` directories/files. I also made some slight changes to the `docs/fields-*` files due to a change in the naming convention on the histology template.

I am open to feedback on my approach here, as this is a bit of a compromise between a) automated testing via Portal CI / running the `./test.sh` suite and b) tests that include live API calls, which I feel are super useful when making sure the CEDAR metadata validation piece is working properly. I am also not the very best at bash scripting so I am sure my `tests-manual/test-dataset-examples-cedar.sh` script has inefficiencies since I based it on the  `tests/test-dataset-examples.sh` file.


src/ingest_validation_tools/upload.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         offline: bool = False,
         ignore_deprecation: bool = False,
         extra_parameters: Union[dict, None] = None,
-        token: str = "",
+        airflow_token: str = "",
```
jpuerto-psc, line 62: Not sure - but it might make most sense to call this globus_token or groups_token since that's really what the auth_tok is.

jpuerto-psc (approved):

jpuerto-psc:
@sunset666 - can you review this please? @gesinaphillips - please resolve the merge conflicts here

gesinaphillips:
@jpuerto-psc Will do! When I went to merge and resolve conflicts earlier it was going to add a bunch of files, I decided to hold off so that those changes didn't make the PR difficult to review. Is that the wrong approach / does it sound like I am merging/rebasing in a dumb way? Regardless, I will get that done.

PR #1231, opened 2023-10-19
Nihdev 407 conditional directory schema validation

jpuerto-psc:
Adds conditional directory support


jpuerto-psc:
@j-uranic - when you have a moment, can you please review the histology directory schema and make sure it is as expected?

jpuerto-psc:
Still need to add documentation changes

sunset666 (approved):

src/ingest_validation_tools/directory_validator.py
```diff
@@ -72,10 +114,24 @@ def _get_required_allowed(dir_schema: List[Dict]) -> Tuple[List[str], List[str]]
     (['this_is_required'], ['this_is_required', 'this_is_optional'])
 
     """
-    allowed = [item["pattern"] for item in dir_schema]
+    allowed = [item["pattern"] for item in dir_schema if "dependency" not in item]
```
gesinaphillips, line 117: Are we weeding out sub-dependencies here? If so, are those not a concern?

gesinaphillips (approved):

src/ingest_validation_tools/directory_validator.py
jpuerto-psc: I don't think that we have sub-dependency situations as of yet. If we do run into that, then we will have to reconsider how we approach this.

PR #1240, opened 2023-10-31
testing fix for finding dir schema

gesinaphillips:
This code was making a bad assumption (that all non-CEDAR schema versions were v0 or v1). This fix passes an is_cedar arg down to `validation_utils > _get_directory_schema_version` so that the determination of CEDAR vs. not-CEDAR is smarter. This still assumes that dir-schema versions are v0 (unless there is a hint in the `extras` directory) for non-CEDAR datasets and v2 (same caveat about hints) for CEDAR datasets. This is imprecise and asking for trouble and should be fixed with the soft assay types integration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Bugfix datetime constraint in library_creation_date.yaml
 - Update LCMS and add NanoSplits
 - Update descriptions for segmentation masks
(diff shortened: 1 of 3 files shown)


sunset666 (approved):

gesinaphillips:
No longer necessary due to soft assay work #1245

PR #1245, opened 2023-11-09
Soft assay types initial integration

gesinaphillips:
This PR includes the initial integration of soft assay types via the [assayclassifier](https://github.com/hubmapconsortium/ingest-api/tree/welling/soft_assay_types_start/src/routes/assayclassifier) Joel is working on. A brief description of functionality:
- SchemaVersion has been changed from a namedtuple to a dataclass. It can now hold loads of info that can be referenced as properties of the class.
- While compiling `effective_tsvs`, the upload makes a call to the assayclassifier for each TSV. This retrieves the canonical assay name and the directory schema string.
- This removes the need for is_cedar `table-schemas` and simplifies the logic for finding the right directory schema*.
     - It doesn't look much simpler because we need to support legacy uploads, some of which might have hints in the `extras` dir. This could probably use a refactor, and will absolutely need it once we're done with non-CEDAR validation.

Many, many of the changes in this PR pertain to testing. So much now runs online that automated testing required some rethinking. 
- I have added mock files for examples that would call out to the assayclassifier (which is most of them) with a mocked response. The code checks the arguments that would have been passed to the assayclassifier in online mode, and raises an exception if they don't match what's expected. If the args match, it returns a mocked response.
- CEDAR validation and URL checking are similarly difficult due to the need for an online connection. This is currently being solved with the addition of a README_ONLINE.md file that is checked when tests are run manually in online mode, and the use of the README.md file (which is missing e.g. URL checking and CEDAR validation results) for offline mode. This should probably be switched to a mocked response as CEDAR becomes the dominant paradigm. 

Also I don't know why I thought we needed the CEDAR API key, we don't for the spreadsheet validator (OOPS). So I removed it.


jpuerto-psc:
> This removes the need for is_cedar table-schemas and simplifies the logic for finding the right directory schema

We do currently need this for documentation purposes still.

gesinaphillips:
> > This removes the need for is_cedar table-schemas and simplifies the logic for finding the right directory schema
> 
> We do currently need this for documentation purposes still.

Whoops so we do. Note updated to read:

"This removes the need to reference is_cedar table-schemas during validation..."

src/ingest_validation_tools/validation_utils.py
```diff
@@ -30,96 +34,167 @@ def dict_reader_wrapper(path, encoding: str) -> list:
...
+    row: Dict,
+    globus_token: str,
+    path: Path,
+    env: str = "dev",
+    offline: bool = False,
+) -> Dict:
+    if "version" in row.keys():
+        row["version"] = int(row["version"])
+    if offline or not globus_token:
+        return mock_response(path, row)
```
jpuerto-psc, line 150: We should rethink this for future iterations. Put a TODO here so that we can discuss this once this push is over.

jpuerto-psc (approved):

src/ingest_validation_tools/validation_utils.py
gesinaphillips: Done!

src/ingest_validation_tools/upload.py
```diff
@@ -73,23 +69,66 @@ def __init__(
         self.errors = {}
         self.effective_tsv_paths = {}
         self.extra_parameters = extra_parameters if extra_parameters else {}
-        self.auth_tok = globus_token
-        self.cedar_api_key = cedar_api_key
+        self.globus_token = globus_token
```
sunset666, line 72: Just a reminder we need to update ingest-pipeline with the new parameters

src/ingest_validation_tools/validation_utils.py
```diff
@@ -249,65 +319,20 @@ def get_tsv_errors(
...
-) -> dict:
-    others = get_other_names()
-    if schema_name in others:
-        schema = get_other_schema(schema_name, version, offline=offline)
-    else:
-        schema = get_table_schema(
-            schema_name, version, offline=offline, optional_fields=optional_fields
-        )
-    return schema
+    # TODO: this is weird, because we're creating an upload for a single file...maybe subclass?
```
sunset666, line 326: Following up with the TODO, maybe this whole routine can go away since it was meant to be used as metadata.tsv stand-alone validator, and that happens on CEDAR?

sunset666 (approved):
Nothing serious here, just small things to consider.

src/ingest_validation_tools/validation_utils.py
```diff
@@ -30,96 +34,168 @@ def dict_reader_wrapper(path, encoding: str) -> list:
...
-        assert "metadata_schema_id" in rows[0]
-        cedar_validation = True
-    except AssertionError:
-        cedar_validation = False
-    return get_table_schema_version_from_row(
-        path, rows[0], cedar_validation=cedar_validation
+        rows = read_rows(path, encoding)
+    except TSVError as e:
+        raise PreflightError(e.errors)
+    assay_type_data = {}
```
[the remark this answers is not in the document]
gesinaphillips: Good catch--removed

src/ingest_validation_tools/validation_utils.py
gesinaphillips: `get_tsv_errors` is called from Ingest API, so I preserved functionality here. It might be worth looking into whether we can change that though (perhaps outside of the scope of this PR?).

src/ingest_validation_tools/upload.py
gesinaphillips: I think this will address the changed params but let me know if you see anything missing https://github.com/hubmapconsortium/ingest-pipeline/pull/792

yuanzhou:
11/29/2023, downstream changes made in ingest-api via PR https://github.com/hubmapconsortium/ingest-api/pull/419

PR #1267, opened 2023-12-04
Multi-assay support

gesinaphillips:
**⚠️ ⚠️ ⚠️ Once again please ignore the entire `examples` dir. Files changed without added testing data: 5. ⚠️ ⚠️ ⚠️** 

This PR introduces multi-assay checking to a) ensure that a multi-assay upload has a valid structure and if so, b) check dir structures and multi-references in a multi-assay-friendly way.

### Functionality ###
After data is retrieved from the assayclassifier endpoint, `Upload._check_multi_assay` runs to do the following:
- Determine whether any schemas include "must_contain" -- if not, this is not a multi-assay upload, so `_check_multi_assay` bails
- Gather up any remaining schemas and treat them as child dataset types to the parent multi-assay type
- Check that all required children are present and that no other, not-allowed types are present
- Check data_paths in parent TSV to ensure that each one is present in all required child TSVs
- If everything is good structure-wise, use the compiled `multi_assay_data_paths` dict to create a separate route for dir schema and multi-ref checking downstream
  - The dedicated multi-assay dir checking ensures that multi-assay shared data paths are checked according to the parent dir schema while any child-only paths in that same upload are checked according to their own dir schema (see `examples/dataset-examples/bad-cedar-multi-assay-visium-with-standalone-histology-bad-dir-schema`)
  - Ref checking will usually log an error if the same data_path is referenced in multiple TSVs (see `examples/dataset-examples/bad-mixed`), but will ignore any shared paths in `multi_assay_data_paths` if they have a "parent" value


src/ingest_validation_tools/upload.py
```diff
@@ -278,38 +255,58 @@ def _get_local_tsv_errors(self) -> Optional[Dict]:
...
+            return self._multi_assay_dir_check(parent[0], data_path, path)
+        # Validate against component structure otherwise
+        elif dataset_types.get("components"):
+            errors = {}
+            for component in dataset_types["components"]:
+                errors.update(self._multi_assay_dir_check(component, data_path, path))
+            return errors
+
+    def _multi_assay_dir_check(
+        self, schema: SchemaVersion, ref_path: Path, data_path: str
```
jpuerto-psc, line 295: I'm a bit confused on the invocation on line 291 and the declaration here, data_path -> ref_path and path -> data_path

src/ingest_validation_tools/upload.py
```diff
@@ -370,14 +367,11 @@ def _validate(
...
         no_ref_errors = self.__get_no_ref_errors()
-        try:
-            multi_ref_errors = self.__get_multi_ref_errors()
-            if no_ref_errors:
-                errors["No References"] = no_ref_errors
-            if multi_ref_errors:
-                errors["Multiple References"] = multi_ref_errors
-        except UnicodeDecodeError as e:
-            errors["Decode Error"] = get_context_of_decode_error(e)
+        multi_ref_errors = self.__get_multi_ref_errors()
```
jpuerto-psc, line 370: Why did you take this out of the try catch block?

src/ingest_validation_tools/upload.py
```diff
@@ -422,6 +416,88 @@ def api_validation(
...
+        self, parent: SchemaVersion, components: List[SchemaVersion]
+    ):
+        """
+        Iterate through child dataset types, check that they are valid
+        components of parent multi-assay type and that no components are missing
+        """
+        not_allowed = []
+        necessary = parent.contains
+        for sv in components:
+            if sv.dataset_type.lower() not in parent.contains:
```
jpuerto-psc, line 447: I think this flow is why we're limited to one multi-assay dataset type per upload, right? Does this also make it such that we can't have a multi-assay dataset and an unrelated single-assay dataset that is **not** of the same dataset type as that multi-assay dataset's components?

src/ingest_validation_tools/upload.py
```diff
@@ -212,29 +168,50 @@ def get_errors(self, **kwargs) -> dict:
...
+        # This is not recursive, so if there are nested multi-assay types it will not work
+        self.multi_assay_data_paths: DefaultDict[
+            str, DefaultDict[str, List[SchemaVersion]]
+        ] = defaultdict(lambda: defaultdict(list))
+        multi_assay_parents = [
+            sv for sv in self.effective_tsv_paths.values() if sv.contains
+        ]
+        if len(multi_assay_parents) == 0:
+            return {}
+        if len(multi_assay_parents) > 1:
```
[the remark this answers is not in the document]
gesinaphillips: I had thought we might only want one parent, but now that you point it out I can see there being an upload with multiple multi-assays. I'll rework this (now if it's easy, later if it'll delay the initial release).

src/ingest_validation_tools/upload.py
gesinaphillips: Ah that's an oversight, you're right. I will see if I can expand the acceptable cases to accommodate non-parent/non-component inclusions in an upload including a multi-assay.

src/ingest_validation_tools/upload.py
gesinaphillips: Oh you're right that's terrible.

gesinaphillips: Made a fix, let me know if that's better!

src/ingest_validation_tools/upload.py
gesinaphillips: Just double checked and yep that's why! Storing the rows on the SchemaVersion helps to avoid reading the TSV rows over and over.

src/ingest_validation_tools/validation_utils.py
```diff
@@ -55,19 +55,23 @@ def get_schema_version(
...
+            f"Assay data not retrieved from assayclassifier endpoint for TSV {path}."
+        )
+        if "assay_type" not in rows[0] and "dataset_type" not in rows[0]:
             message += ' Does not contain "assay_type" or "dataset_type".'
-        elif "assay_type" in rows[0].keys():
+        elif "assay_type" in rows[0]:
             message += f' Assay type: {rows[0].get("assay_type")}.'
-        elif "dataset_type" in rows[0].keys():
+        elif "dataset_type" in rows[0]:
             message += f' Dataset type: {rows[0].get("dataset_type")}.'
```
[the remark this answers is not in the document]
gesinaphillips: Good catch, I forgot to remove the check on line 70 when I moved the initial message out of this block. Fixed!

PR #1285, opened 2024-01-12
Prevent repeated processing of files/contributors.tsv

gesinaphillips:
This PR updates validation to prevent repeated validation of contributors.tsv (via the Spreadsheet Validator API) and individual files during plugin checking. This should reduce the time it takes to validate large multi-assay uploads.

## Files to look at ##
- src/ingest_validation_tools/upload.py
- src/ingest_validation_tools/plugin_validator.py

## Files you can ignore ##
Example files were updated in two ways:
- running online tests updated all responses due to changes to assayclassifier responses (addition of `contains-pii` and `primary`)
- slight changes to formatting required regeneration of README files.
- addition of manual tests for plugins, because plugins are turned off for automated testing
     - `tests-manual/` has also been updated to add a script for plugin testing

## Details ##
- contributors.tsv: all unique contributor_paths in a given metadata.tsv are collected and each unique path is validated once. In practice, this should result in the contributors.tsv being sent to the Spreadsheet Validator once per metadata.tsv (rather than once per row per TSV).
- plugin validation:
     - this runs as normal for non-multi-assay uploads.
     - For multi-assay uploads, whatever metadata.tsv is identified as the parent type is used as a manifest for file checking, so each data dir is only checked once per upload.
          - That is, only that single TSV is sent for plugin validation (rather than each TSV being sent and therefore each data dir being validated _n_ times). 
          - The plugins do have some sense of what dataset type they are running on, but this seems irrelevant because currently the only types that are used are CODEX and Publications (neither of which are related to multi-assay). Bare-bones infrastructure has been added to support sending component dataset types to plugins.


src/ingest_validation_tools/plugin_validator.py
```diff
@@ -155,6 +161,6 @@ def validation_error_iter(
     if not base_dir.is_dir():
         raise ValidatorError(f"{base_dir} should be the base directory of a dataset")
     for cls in validation_class_iter(plugin_dir):
-        validator = cls(base_dir, assay_type)
+        validator = cls(base_dir, assay_type, contains)
```
jpuerto-psc, line 164: As long as none of the existing Plugins override the __init__ class defined on line 46 then this should be fine. Just making sure that that is the case.

jpuerto-psc (approved):

src/ingest_validation_tools/plugin_validator.py
gesinaphillips: Good point, I will double check that.

gesinaphillips: All good with this at present.

sunset666 (approved):

PR #1290, opened 2024-01-25
Jpuerto/nihdev 454 ivt adjust dir schema validation for shared uploads

jpuerto-psc:
This PR adds support for the `global` and `non_global` shared uploads defined here: https://docs.google.com/document/d/1n2McSs9geA9Eli4QWQaB3c9R3wo5d5U1Xd57DWQfN5Q

This PR also adds an example shared PhenoCycler dataset for automated testing.


jpuerto-psc:
Two example uploads in DEV that passed metadata/directory schema validation (fail on plugins because the files are empty).

gesinaphillips (approved):

gesinaphillips (approved):

jpuerto-psc:
Leaving this open for now as it requires some input from Stephen Fisher

src/ingest_validation_tools/upload.py
```diff
@@ -56,6 +56,7 @@ def __init__(
         extra_parameters: Union[dict, None] = None,
         globus_token: str = "",
         run_plugins: bool = False,
+        shared_directories: Union[set, None] = False,
```
gesinaphillips, line 59: Typing is wrong here, and this is not accessed--could be removed?

src/ingest_validation_tools/upload.py
```diff
@@ -752,6 +772,30 @@ def __get_no_ref_errors(self) -> dict:
...
+                for full_path_row_non_global_file, rel_path_row_non_global_file in \
+                        row_non_global_files.items():
+                    if not full_path_row_non_global_file.exists():
+                        errors[",".join(row_references)] = \
+                            f"{rel_path_row_non_global_file} not exist in upload."
+        else:
+            # Catch case 2
+            if self.shared_directories:
+                errors["Upload Errors"] = "No non_global_files specified but " \
+                                          "upload has global & non_global directories"
```
gesinaphillips, line 795: Very nitpicky: I think this would trigger if an upload had a global _and/or_ non_global directory, rather than needing to have both ("&"). Suggest slight update from `&` -> `and/or`

src/ingest_validation_tools/directory_validator.py
```diff
@@ -22,27 +22,30 @@ def validate_directory(
         required_patterns, allowed_patterns = _get_required_allowed(schema_files)
         dependencies = _get_dependencies(schema_files)
     except KeyError as e:
-        raise DirectoryValidationErrors(f"Error finding patterns for {path}: {e}")
+        raise DirectoryValidationErrors(f"Error finding patterns for {','.join(paths)}: {e}")
```
gesinaphillips, line 25: I think this will break since it's trying to join PosixPath objects rather than strings. Maybe something like `f"Error finding patterns for {','.join([str(path) for path in paths])}: {e}"`?

gesinaphillips:
Unless uploads with shared directories will be passed "global" and "non_global" as part of the Upload's `upload_ignore_globs` argument, I think the `__get_no_ref_errors` method needs to be updated to ignore the presence of global/non_global directories because they are not technically referenced. If you move the new `good-cedar-phenocycler-shared` example from `dataset-iec-examples` to `dataset-examples` (because the two dirs are called with different args during testing) you should see the following error when you run the tests:

```
Reference Errors:
  No References:
    Files:
    - global.
    - non_global.
```

One way of solving this would be to add to self.upload_ignore_globs if self.shared_directories is True, e.g.:

```
        if self.shared_directories:
            self.upload_ignore_globs = [
                *self.upload_ignore_globs,
                *["global", "non_global"],
            ]
```

I added this inside `_get_no_ref_errors` just before `non_metadata_paths` is created and it avoided the no-ref error in the phenocycler example. HOWEVER I am not sure if this has unwanted downstream effects!

jpuerto-psc:
> Unless uploads with shared directories will be passed "global" and "non_global" as part of the Upload's `upload_ignore_globs` argument, I think the `__get_no_ref_errors` method needs to be updated to ignore the presence of global/non_global directories because they are not technically referenced. If you copy the new `good-cedar-phenocycler-shared` example from `dataset-iec-examples` to `dataset-examples` (because the two dirs are called with different args during testing) you should see the following error when you run the tests:

@gesinaphillips - we should be fine here, the ingest-pipeline invocation basically ignores all directories at the top-level (which is why `extras` doesn't get caught either).

src/ingest_validation_tools/directory_validator.py
jpuerto-psc: Thanks for pointing this out - I knew I had missed one of these when I was fixing the error messages!

PR #1303, opened 2024-02-26
Testing fix

gesinaphillips:
## Once again, most changes are to example files!

This PR removes any testing logic from production code and moves it purely into the testing suite. It maintains functionality of automatic and manual testing suites.

Changelog (bolded portion constitutes code in most need of review):
- **Addition of `tests/test_dataset_examples.py` to add automated online testing that is reliant on fixture data and, as before, README.md files.**
- **Addition of `tests-manual/update_test_data.py` as a command line tool to (hopefully easily) update fixture data and README files in future.**
- Removal of `src/ingest_validation_tools/test_validation_utils.py` and any referenced functions in `validation_utils.py`.
- Removal of `examples/**/MOCK_RESPONSE.json` and `examples/**/README_ONLINE.md` files.
- Addition of `examples/**/fixtures.json` files.
- Slight changes to `src/ingest_validation_tools` > `plugin_validator.py`, `schema_loader.py`, and `upload.py` to control verbosity and streamline patching for testing method. 
- Updates to other files in `tests/` and `tests-manual/` to maintain compatibility.
- Updates to documentation in `tests-manual/README.md`.

## Notes
- I had a little trouble parameterizing the tests because I wanted to maintain compatibility with the `test.sh` runner. Any insight is appreciated as I know it's a little messy.
- I chose to go with individual fixtures files because I find that associating fixture data with a given upload dir is helpful for debugging. Let me know if you disagree.


tests-manual/update_test_data.py
```diff
@@ -0,0 +1,300 @@
...
+        errors = self.upload.get_errors()
+        report = ErrorReport(info=info, errors=errors)
+        if "Too many requests" in report.as_md():
+            raise Exception(
+                f"Something went wrong with Spreadsheet Validator request for {self.dir}."
+            )
+        if "fixtures" not in self.exclude:
+            new_data = self.update_fixtures(report)
+            fixtures = self.open_or_create_fixtures()
+            diff = DeepDiff(fixtures, new_data, ignore_order=True, report_repetition=True)
```
gesinaphillips, line 71: @jpuerto-psc So it turns out the `==` was doing a pretty thorough job with nested dicts, but it would absolutely get tripped up if there was any difference in list ordering. I think list ordering is not meaningful here so I have set it to `False` (while making sure to catch any duplicates in the list, which are ignored by default if you ignore order). This also lets us get a better diff output, which I think is nice. What do you think?

jpuerto-psc: That makes sense to me! I agree that list ordering is not relevant here and that set equivalence is more important.

jpuerto-psc (approved):

sunset666 (approved):

PR #1312, opened 2024-03-18
Dir schema major/minor versions

gesinaphillips:
Attempt at accommodating major/minor directory schema versioning. Addressed @jswelling's comments about sorting--thank you!

~~CANNOT BE MERGED UNTIL CHANGES TO ASSAYCLASSIFIER ARE MADE. "codex-v1-with-dataset-json" needs to be changed to codex-v1.~~

Changes:
- Most relevant changes are to `schema_loader.py`, `upload.py`, and `validation_utils.py`. 
- Added tests for dir schema minor version to `tests/test_dataset_examples.py`.
     - Other changes in `tests/` and `tests-manual/` are bugfixes/additional functionality.
- Switched existing dir schemas to new format and re-ran tests (also includes some updates to the assayclassifier).

What is not done:
- (Future phase) Save minor version to upload metadata where called in ingest-pipeline.
- (Future potential phase) Check for previous run of directory schema validation and plugin validation and do not re-run if previously valid.
     - Might want to re-run anyway, perhaps make configurable.


src/ingest_validation_tools/schema_loader.py
```diff
@@ -260,6 +260,19 @@ def get_directory_schema(
     return schema
 
 
+def get_possible_directory_schemas(dir_schema: str) -> Optional[Dict]:
+    schemas = {}
+    # this assumes that versions are numbered starting at x.0, no whole numbers
+    directory_schema_minor_versions = _directory_schemas_path.glob(f"{dir_schema}*.yaml")
```
jswelling, line 266: If dir_schema is 'codex-v1', won't that wildcard match 'codex-v17' ?  That is, don't you need an explicit '.' ?

src/ingest_validation_tools/validation_utils.py
```diff
@@ -156,41 +156,54 @@ def get_data_dir_errors(
...
-    schema_warning = (
-        {f"{schema_warning_fields[0].title()} directory schema": dir_schema}
-        if schema_warning_fields
-        else None
-    )
+    # Collect errors, discard if schema validates against a minor version
+    errors = []
+
+    # Make sure possible_schemas is sorted by key (descending) to evaluate highest minor version first
+    for schema_name, schema in sorted(possible_schemas.items(), reverse=True):
```
jswelling, line 172: In what order does this sort put ['codex-v1.14', 'codex-v1.3'] ?

src/generate_docs.py
```diff
@@ -77,7 +77,10 @@ def main():
             current["metadata"][v] = schema
 
     for v, schema in directory_schemas.items():
-        if not v.isdigit() or int(v) < 2:
+        try:
+            a = float(v)
```
sunset666, line 81: If there is no use for 'a' down the code, I believe it could be just assert float(v).

sunset666: Also could it fail because v isn't a number? maybe another except will be needed

gesinaphillips: Good catch(es)! Updated, let me know what you think. We don't seem to have cared about v not being a number before so I am just catching that ValueError in the same spot.

sunset666 (approved):

jpuerto-psc (approved):

PR #1319, opened 2024-04-09
Convert upload errors from dict to dataclass

gesinaphillips:
Change to support writing to and pulling from `Upload.errors` without having to know specific key names. 

Anticipated benefits:
- Makes output configuration much more flexible.
- Supports error classification ([Jira ticket NIHDEV-504](https://pscdevteam.atlassian.net/browse/NIHDEV-504)) using the ErrorDict object to access specific error types as attributes rather than by string matching.
- Will support getting dir schema version ([Jira ticket NIHDEV-534](https://pscdevteam.atlassian.net/browse/NIHDEV-534)) from the InfoDict object without having to dig (have not yet created `Upload.dir_schema` attr yet).
- I am not positive it's more intuitive to add things to the error output now but at least now it's typed so you can get some hints about what data types are expected?

Guide to changes:
- Major changes in `upload.py` and `error_report.py`.
- Major changes to testing in `tests/test_dataset_examples.py` and to programmatic updating functionalities in `tests-manual/update_test_data.py`.
- `examples/` dir contains simplified `README.md` files ([Jira ticket NIHDEV-522](https://pscdevteam.atlassian.net/browse/NIHDEV-522)) and updated fixtures.
- Output of `get_tsv_errors` has been adjusted in conversation with @maxsibilla.

Todo:
- Discuss `test-tsv-examples.py`--either needs to be fixed or deprecated, I suggest deprecation (see [this commit](https://github.com/hubmapconsortium/ingest-validation-tools/commit/7f3579e58232038e3f575881c49aa840b5728dec) where I tested it out). We could add some unit tests that test `get_tsv_errors` instead.
- If this approach seems sound, this is what the error classification looks like in ingest-pipeline: hubmapconsortium/ingest-pipeline#861
- Adjustment of Upload to potentially be the source of truth for more information about the upload (e.g. dir schema, main assay type), rather that burying this info in Upload.effective_tsv_paths.


src/ingest_validation_tools/validation_utils.py
```diff
@@ -176,7 +173,10 @@ def get_data_dir_errors(
     if {x.name for x in shared_directories} == expected_shared_directories:
         # If they exist create a list of the paths
         data_paths = list(shared_directories)
-    # Otherwise, do nothing we can just use the predefine data_path
+    # Otherwise, do nothing we can just use the predefined data_path
+
+    for data_path in data_paths:
+        assert Path(data_path).exists, f"data_path {data_path} does not exist!"
```
jpuerto-psc, line 179: same question RE assertion

src/ingest_validation_tools/upload.py
```diff
@@ -106,124 +85,140 @@ def __init__(
...
-                logging.info("Running plugin validation...")
+        # Collect errors
+        self._get_local_tsv_errors()
+        self._get_directory_errors()
+        self.validation_routine()
+        self._get_reference_errors()
+
+        # Plugin error checking is costly, by default this bails
+        # if other errors have been found already
+        if self.errors and not self.run_plugins:
```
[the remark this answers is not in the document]
gesinaphillips: To answer your first question, yes. I'm thinking now that this isn't necessary, because the default behavior you describe is sensible. I'll test and make sure.

src/ingest_validation_tools/upload.py
```diff
@@ -106,124 +85,140 @@ def __init__(
...
+            for path in (tsv_paths if tsv_paths else self.directory_path.glob(f"*{TSV_SUFFIX}"))
+        }
+
+        self.effective_tsv_paths = {
+            k: unsorted_effective_tsv_paths[k] for k in sorted(unsorted_effective_tsv_paths.keys())
+        }
+
     def _check_single_assay(self):
+        # TODO: is this necessary?
+        # Is there a case where there should be more than one effective_tsv_path for a non-multi assay upload?
```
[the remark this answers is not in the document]
gesinaphillips: Good point about placement of the note, I'm just moving it to a different spot for later consideration.

src/ingest_validation_tools/validation_utils.py
gesinaphillips: This one is caught where `get_data_dir_errors` is called in upload.py (inside try block in _check_data_path).

src/ingest_validation_tools/upload.py
```diff
@@ -235,133 +230,128 @@ def _check_single_assay(self):
...
+                errors = self._get_ref_errors(ref, schema, path)
+                if errors:
+                    self.errors.upload_metadata.update(errors)
 
-    def _get_directory_errors(self) -> dict:
-        errors = {}
+    def _get_directory_errors(self):
         if self.is_multi_assay and self.multi_parent:
             for data_path in self.multi_assay_data_paths:
+                assert Path(data_path).exists, f"Shared data path {data_path} does not exist!"
```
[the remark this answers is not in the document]
gesinaphillips: Went through and fixed asserts, I decided to just let the eventual call to `validate_directory` handle missing paths instead.

src/ingest_validation_tools/upload.py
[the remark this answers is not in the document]
gesinaphillips: Seems to work fine!

src/ingest_validation_tools/upload.py
```diff
@@ -106,124 +85,140 @@ def __init__(
...
+        return self.errors
 
-        return errors
+    def get_app_context(self, submitted_app_context: Dict):
+        default_app_context = {
+            "entities_url": "https://entity.api.hubmapconsortium.org/entities/",
+            "ingest_url": "https://ingest.api.hubmapconsortium.org/",
+            "request_header": {"X-Hubmap-Application": "ingest-pipeline"},
+        }
+        if {*default_app_context} - {*submitted_app_context}:
```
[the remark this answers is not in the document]
gesinaphillips: Tested out fine.

src/ingest_validation_tools/validation_utils.py
gesinaphillips: Actually this didn't make the cut either, removed.

src/ingest_validation_tools/upload.py
```diff
@@ -146,28 +144,28 @@ def get_errors(self, **kwargs) -> ErrorDict:
...
-        # Plugin error checking is costly, by default this bails
+        # Plugin error checking is costly, this bails
         # if other errors have been found already
-        if self.errors and not self.run_plugins:
+        if self.errors:
             self.errors.plugin_skip = (
                 "Skipping plugins validation: errors in upload metadata or dir structure."
             )
-        elif self.run_plugins:
+        else:
```
jpuerto-psc, line 153: @gesinaphillips - I think we only want to run plugin validation if the self.run_plugins is true. I think the issue with the previous logic was that if the self.run_plugins is true it wouldn't fall into the self.errors if statement. So I think the resolution here is to bring the `elif self.run_plugins` back and leave the `and not self.run_plugins` off the `self.errors` check.

gesinaphillips: Oh I think I see the reason for my confusion. My initial implementation of `run_plugins` was meant to let you run plugin validation _even if_ there were upstream errors (I think the idea was that there might be cases where you want to get ALL errors). However given the size of most uploads that is probably not needed. 

I believe this change (fail if errors, skip plugins unless explicitly passed `run_plugins=True`) would require us to either: set `run_plugins` to default True or add a run_plugins arg to the validate_upload.py DAG (and maybe others?).

gesinaphillips: I guess this purpose makes the original `elif self.run_plugins` redundant at best, admittedly.

gesinaphillips: ANYWAY decision point:
- Leave as-is (change `elif self.run_plugins` to just `else` I think): default logic is to set `run_plugins` to False so it fails on any errors, and only run plugins if no errors OR if run_plugins explicitly set to True. Can run plugins if there are errors but requires passing `run_plugins=True`.
- Change A: change `if self.errors and not self.run_plugins` to just `if self.errors`: default logic is that run always fails if there are errors, and plugins only run if `run_plugins` is explicitly set to True. Change needed in ingest-pipeline.
- Change B: change `if self.errors and not self.run_plugins` to just `if self.errors` and default parameter to `run_plugins=True`: default logic is that run always fails if there are errors, but plugins always run unless there are errors (and unless `run_plugins` is explicitly set to False). No change to ingest-pipeline.

jpuerto-psc (approved):

PR #1321, opened 2024-04-15
Documentation update

gesinaphillips:
This PR does two main things:
- Deprecates field files that were being automatically generated based on previous gen YAML table schema files.
     - ~~I was really not sure how we want to do this since I know we want to keep the files rather than deleting. I chose to go with in-place `_deprecated` directories, but an alternative is to have a top-level `_deprecated` directory with sub-dirs that mirror the original locations of the deprecated files. E.g. `docs/field-assays.yaml` -> `_deprecated/docs/field-assays.yaml`. This would consolidate things better but may be somewhat confusing as a directory.~~ 
     - ~~Or we could just dump them all in one folder.~~
     - ~~Anyway let me know what you think.~~
     - Went with top-level `_deprecated` directory.
- Updates documentation to be more correct.
     - Many out-of-date sections revised.
     - Replaced `offline` parameter on Upload with `no_url_checks` as there is no longer a true offline validation path. However, previous gen validation tries to run online URL checks unless they are explicitly turned off, so this is necessary for testing. This change also allows for user-facing documentation for `validate_upload.py` to hand-wave the portion about obtaining a Globus token, as that functionality can be turned off (this does of course mean that a user could validate locally and still have URL check errors when they submit their upload, but that seems like a small trade-off considering the currently busted status). This update allows for a (potentially temporary) state of correct user-facing documentation for how to use `validate_upload.py` to validate an upload's directory schema.


jpuerto-psc:
@gesinaphillips I think I prefer the method of having a single `_deprecated` directory at the top level and having all deprecated files live underneath that in the same directory tree structure that they currently live in. Having `_deprecated` all throughout the project makes it a bit confusing to read through.

gesinaphillips:
> @gesinaphillips I think I prefer the method of having a single `_deprecated` directory at the top level and having all deprecated files live underneath that in the same directory tree structure that they currently live in. Having `_deprecated` all throughout the project makes it a bit confusing to read through. Once you've made that change, I'm happy to approve.

Having slept on this a bit, I agree @jpuerto-psc. On it!

jpuerto-psc (approved):

sunset666 (approved):

PR #1323, opened 2024-04-18
Add 'version' to plugin base class

jswelling:
Add a 'version' field to the base class for plugin tests, to hold a semantic version string


gesinaphillips (approved):
Just need to update the changelog!

PR #1329, opened 2024-05-07
Jpuerto/donor yaml update

jpuerto-psc:


jpuerto-psc:
@maxsibilla - I would recommend that on any future syncs between HuBMAP/SenNet for this repository that you **do not** pull in any changes related to these YAML files.

maxsibilla:
@jpuerto-psc Removing the fields and files related to murine source was something we were about to do so thank you. There shouldn't be anymore syncing moving forward, the SenNet IVT repo is now defunct and with the addition of the new murine CEDAR spec we no longer accept any format of metadata that isn't CEDAR.

PR #1330, opened 2024-05-13
Entity type constraints

gesinaphillips:
Adds entity constraint checking support to validation. The bare constraints URL (e.g. `https://entity.api.sennetconsortium.org/constraints/`) must be passed as part of the `Upload.app_context` param for this to run as the endpoint is not currently implemented in HuBMAP. 

This should not require any updates in HuBMAP at this time, and should not noticeably affect HuBMAP's validation process.

## tl;dr ##
- Major changes are in `src/ingest-validation-tools` > `enums.py`, `upload.py`, and `validation_utils.py`.

## Major functionality ##
- Adds enums for entity types, plus some tooling since there are different ways to refer to types ("sample" vs. "sample-block" vs. "block") and there are different categories of types ("dataset" vs. "others").
     - Some existing logic probably remains that could be switched over to these enums.
- Adds entity type data about the entity the TSV represents while it's creating the schema, stores on schema object to be used later (`SchemaVersion.entity_type_info`).
     - In the case of samples this currently requires an additional entity-api call because there's not a straightforward way to get the sample type (block, suspension, section, organ) out of the metadata TSV without either string comparing column headers or CEDAR IDs, both of which are fragile.  
- Gets entity-api data about each ancestor ID checked (`source_id`, `parent_sample_id`) during existing URL checking, and stores formatted response data on the schema object (`SchemaVersion.ancestor_entities`).
- After URL checking, sends a payload of ancestor-descendant pairs (representing `ancestor ID in TSV` : `entity info about TSV descendant`) to the constraints endpoint, and formats any errors that are returned.
- Testing added.


maxsibilla (approved):
Verified this branch works with the SenNet entity constraints endpoint.

sunset666 (approved):

PR #1339, opened 2024-06-25
J uranic dirschema visium updates

j-uranic:
Add optional .json files in  lab_processed/images/ directory for Visium with probes and Visium no probes per Stephen Fisher.


j-uranic:
CANCELED - updates made with new versioning process on new branch.

PR #1343, opened 2024-07-02
Add globus_token to get_assaytype_data

gesinaphillips:
- Add globus_token to `validation_utils.get_assaytype_data`
- Update tests
- Switch from concatenating URLs to urllib methods


src/ingest_validation_tools/upload.py
```diff
@@ -371,7 +378,8 @@ def _constraint_checks(self, schema: SchemaVersion):
             "Authorization": f"Bearer {self.globus_token}",
             "Content-Type": "application/json",
         }
-        url = f"{self.app_context['constraints_url']}?match=True&order={CONSTRAINTS_CHECK_METHOD}"
+        params = urlencode({"match": True, "order": CONSTRAINTS_CHECK_METHOD})
+        url = urljoin(self.app_context["constraints_url"], f"?{params}")
         response = requests.post(url, headers=headers, data=data)
```
jpuerto-psc, line 383: You should be able to use the `params` parameter in the post call here:

https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

So:
```
params = {"match": True, "order" CONSTRAINTS_CHECK_METHOD}
requests.post(self.app_context['constraints_url'], headers=headers, data=data, params=params)
```

src/ingest_validation_tools/upload.py
```diff
@@ -627,6 +632,7 @@ def _check_url(
                     *get_entity_type_vals(response.json()),
                 )
         elif field in ["orcid_id", "orcid"]:
+            url = urljoin(constrained_fields[field], f"?q=orcid:{value}")
             headers = {"Accept": "application/json"}
             response = requests.get(url, headers=headers)
```
jpuerto-psc, line 637: Same suggestion as above

src/ingest_validation_tools/upload.py
gesinaphillips: That's so much better!

src/ingest_validation_tools/upload.py
gesinaphillips: The params for the ORCID API are so hateful, that {"q": "orcid:{value}"} pattern threw me until literally right this second. Thank you for helping me give this a second look :)

jpuerto-psc (approved):

PR #1351, opened 2024-08-26
Adjustments for new Plugins

sunset666:
Updated parameters to be used by new plugins (EPICs).


gesinaphillips (approved):
Still looking at the ingest-validation-tests PR but this looks good to me on its own!

gesinaphillips (approved):

PR #1353, opened 2024-08-27
Juranic/add celldive dirschema

j-uranic:
Add next-gen directory schema for Cell DIVE (replaced placeholder text in celldive-v2.0.yaml).
Major change: @gesinaphillips, this is for rules engine update.


jpuerto-psc:
Ty @j-uranic - @gesinaphillips, we should be good here, rules engine already has 2.0 listed as the dir schema version to check against.

jpuerto-psc (approved):

PR #1364, opened 2024-09-26
Testing updates

gesinaphillips:
- Updates test data due to changes in dir schema versions and soft assay type endpoint responses
- Fixes upload `run_plugins` param to actually behave as a boolean
- A few cleanups/bugfixes


src/ingest_validation_tools/upload.py
```diff
@@ -600,13 +607,9 @@ def _get_url_fields(
...
         check = {k: v for k, v in row.items() if k in constrained_fields}
         for check_field, value in check.items():
-            if check_field in self.check_fields and not self.globus_token:
-                raise ErrorDictException(
-                    "No token received to check URL fields against Entity API."
-                )
-            # TODO: could just split if there's a comma in the field
-            elif check_field == "parent_sample_id":
-                url_fields["parent_sample_id"] = value.split(",")
+            # TODO: could just split if there's a comma in the field regardless of field name
```
jpuerto-psc, line 610: I think this is a good approach. I don't think we want to split fields arbitrarily since free text fields could just have commas in them anyway.

sunset666 (approved):

jpuerto-psc (approved):

src/ingest_validation_tools/upload.py
gesinaphillips: Agreed, probably best to be targeted.

PR #1420, opened 2025-06-02
Maxsibilla/other organ type check

maxsibilla:


jpuerto-psc:
@gesinaphillips this change looks good to me, but want to confirm w you before merging in.

PR #1431, opened 2025-07-14
Add Olink directory schema

djbordelon:


gesinaphillips (approved):

djbordelon:
Thanks @gesinaphillips , I thought devs would need to do that table-schema yaml, but it makes sense to me now.

gesinaphillips:
I think the devs are supposed to do the table-schema file. This is a public apology to @djbordelon !!

Note: Proteomics doc category to be created in follow-up PR.

gesinaphillips (approved):

PR #1433, opened 2025-07-14
Modify Phenocycler directory schema

djbordelon:
Per Melissa Farrow, items in the raw temp folder are made optional: "We spoke about the temp folder earlier this spring and arrived at the conclusion that files from that folder would be optional given the storage burden that temp folder incurs."


safisher:
My understanding is that these are required unless the data provider is unable to provide them. We want data providers to provide these, so they are flagged as "required". If it's too complicated from the ingest perspective and blocking data ingest, then this could be made optional.

djbordelon:
In that case, we can continue using our current workaround, which is to transfer in empty placeholder files.

PR #1509, opened 2026-05-14
Plugin kwargs

gesinaphillips:
## Note
This PR is required for update to IVTests but maintains backward compatibility with current ingest-pipeline code (so the upstream work can be merged/released when convenient).

## Updates
- Renamed `Upload.extra_parameters` -> `Upload.plugin_kwargs`.*
- To be removed:  `**kwargs` param from `Upload.get_errors()`. Use `Upload.plugin_kwargs` param to pass plugin-specific kwargs instead.*
- Pass `Upload.plugin_kwargs` down to plugins.

\* Separate PR to `ingest-pipeline` required (see: https://github.com/hubmapconsortium/ingest-pipeline/pull/1278). Param preserved briefly for the switchover. Plugin kwargs will not be passed to plugins until `ingest-pipeline` is updated, but I'm not convinced they were being passed in properly before so I think this is fine.

## Bugfixes
- Ensured uniformity in kwargs passed to plugins between shared/non-shared uploads.
     - Plugins will receive `schema_rows`.
     - Plugins will log the same based on `verbose`.

### DIFF
--- a/src/ingest_validation_tools/plugin_validator.py
+++ b/src/ingest_validation_tools/plugin_validator.py
@@ -109,7 +109,6 @@ def validation_error_iter(
         except Exception as e:
             raise ValidatorError(f"Could not import from plugin_dir {plugin_dir}: {e}")
         for val_class in validation_class_iter():
-            kwargs["verbose"] = verbose
             validator = val_class(
                 paths, assay_type, contains, verbose, schema, globus_token, app_context, **kwargs
             )
--- a/src/ingest_validation_tools/plugin_validator.py
+++ b/src/ingest_validation_tools/plugin_validator.py
@@ -48,7 +48,7 @@ def run_plugin_validators_iter(
     if is_shared_upload:
         paths = [Path(metadata_path).parent / "global", Path(metadata_path).parent / "non_global"]
         for k, v in validation_error_iter(
-            paths, sv.dataset_type, plugin_dir, sv.contains, **kwargs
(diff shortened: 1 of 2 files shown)


src/ingest_validation_tools/plugin_validator.py
```diff
@@ -43,38 +43,41 @@ def run_plugin_validators_iter(
...
             if any(row[column_name] != sv.dataset_type for row in sv.rows):
                 raise ValidatorError(f"{metadata_path} contains more than one assay type")
+    if is_shared_upload:
+        data_paths = [
+            Path(metadata_path).parent / "global",
+            Path(metadata_path).parent / "non_global",
+        ]
+    else:
+        data_paths = get_data_paths_from_tsv(metadata_path, sv)
+    for k, v in validation_error_iter(
```
gesinaphillips, line 53: Ensure that identical args/kwargs (except for `data_path`) are passed for shared/non-shared uploads.

src/ingest_validation_tools/plugin_validator.py
```diff
@@ -43,38 +43,41 @@ def run_plugin_validators_iter(
...
+        verbose=verbose,
+        schema_rows=sv.rows,
+        globus_token=globus_token,
+        app_context=app_context,
+        **kwargs,
+    ):
+        yield k, v
+
 
+def get_data_paths_from_tsv(metadata_path: PathOrStr, sv: SchemaVersion):
```
gesinaphillips, line 67: Logic not modified, just moved into function.

jpuerto-psc (approved):

